PR #4, opened 2018-03-15
add basic i18n package

Hamms:
Adds a basic i18n package that defines a new subclass of mezzanine's Page which should allow for internationalization, with the goal of enabling us to begin uploading strings to crowdin to be translated. Rendering translated content will come later.

Usage is simple; to enable a piece of content to be translated, simply have it inherit from InternationalizablePage rather than from Page. If the piece of content defines extra fields in addition to those provided by Page that should also be translated, extend the `internationalizable_fields` classmethod appropriately.

For example,

```diff
diff --git a/lessons/models.py b/lessons/models.py
index 5fdd44a..f4199bb 100644
--- a/lessons/models.py
+++ b/lessons/models.py
@@ -24,6 +24,7 @@ from documentation.models import Block
 from django_slack import slack_message
 
 from curriculumBuilder import settings
+from i18n.models import InternationalizablePage
 
 from django_cloneable import CloneableMixin
 
@@ -210,7 +211,7 @@ Complete Lesson Page
 """
 
 
-class Lesson(Page, RichText, CloneableMixin):
+class Lesson(InternationalizablePage, RichText, CloneableMixin):
     overview = RichTextField('Lesson Overview')
     short_title = models.CharField('Short Title (optional)', help_text='Used where space is at a premium',
                                    max_length=64, blank=True, null=True)
@@ -244,6 +245,10 @@ class Lesson(Page, RichText, CloneableMixin):
     class Meta:
         ordering = ["number"]
 
+    @classmethod
+    def internationalizable_fields(cls):
+        return super(Lesson, cls).internationalizable_fields() + ['overview']
+
     def __unicode__(self):
         return self.title
```

Some considerations. Obviously, this approach represents a custom implementation rather than using any of the various existing modeltranslation methods. The two major motivations for doing so are:

First that most (all?) of the existing methods are written with the assumption that you ultimately want your site to be browsable in the translated versions; an assumption that does not hold true for us since we simply want our generated static content to be generated in multiple languages.

Second that most (all?) of the existing methods leverage the native django translation format, which is itself built on [gettext](https://www.gnu.org/software/gettext/), a translation format that foregoes using key identifiers for translated strings in lieu of using the content of the source strings themselves as identifiers; a system that does not allow us to take advantage of the full range of translator assistance provided by crowdin.

Josh, any thoughts on this approach? Any major concerns?


mrjoshida:
No concerns here, LGTM.

PR #7, opened 2018-03-15
Fix travis

Hamms:
Lots of iterative commits here, but the ultimate work done boils down to just three changes:

1. Use `manage.py migrate` instead of `manage.py syncdb`
2. Configure correct version of Postgres
3. Squash some early commits that otherwise caused problems

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,9 +1,13 @@
 language: python
+
 python: 2.7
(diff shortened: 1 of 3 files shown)


Hamms:
NP! I was missing my green checkmarks too much not to

PR #11, opened 2018-04-19
extend vocablink to support optional override

Hamms:
Vocablinks can now look like `[v vocabWord][overrideWord]` as well as like the original `[v vocabWord]` syntax. In the new form, the rendered tag would still use the original `vocabWord` and definition for the hover text, but use `overrideWord` for the hoverable element.

This will allow us to do things like this for i18n

image

### DIFF
--- a/curriculumBuilder/vocablinks.py
+++ b/curriculumBuilder/vocablinks.py
@@ -4,13 +4,14 @@ from markdown.inlinepatterns import Pattern
 
 from lessons.models import Vocab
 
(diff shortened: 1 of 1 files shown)


Hamms:
Verified that this change in syntax shouldn't break any existing content by publishing all of CSF and CSP to my local machine both before and after the change and comparing the published files.

Hamms:
I like that! It's totally compatible with the i18n process, too

PR #13, opened 2018-04-26
Make Activities and Chapters internationalizable

Hamms:
Follow-up work: make Activity names unique per lesson and give them a string-based `i18n_key`

### DIFF
--- a/curricula/models.py
+++ b/curricula/models.py
@@ -560,7 +560,7 @@ Unit Chapter
 """
 
 
-class Chapter(Page, RichText, CloneableMixin):
+class Chapter(InternationalizablePage, RichText, CloneableMixin):
     ancestor = models.ForeignKey('self', blank=True, null=True, on_delete=models.SET_NULL)
     number = models.IntegerField('Number', blank=True, null=True)
     questions = RichTextField(blank=True, null=True, help_text="md list of big questions")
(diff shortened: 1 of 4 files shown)


curricula/models.py
```diff
@@ -560,7 +560,7 @@ def clone(self, attrs={}, commit=True, m2m_clone_reverse=True, exclude=[]):
 """
 
 
-class Chapter(Page, RichText, CloneableMixin):
+class Chapter(InternationalizablePage, RichText, CloneableMixin):
```
mrjoshida: Is this the base class to be using going forward for other new content types? In other words, is there any reason not to use InternationalizablePage if we don't actually plan to translate the content any time soon?

lessons/models.py
```diff
@@ -515,7 +515,7 @@ def standards_by_category(self):
 """
 
 
-class Activity(Orderable, CloneableMixin):
+class Activity(Internationalizable, Orderable, CloneableMixin):
```
mrjoshida: Similar question to above. Moving forward is there a cost to making all new things as internationaliable, or should we wait to actually need translation?

curricula/models.py
mrjoshida: Good to know, thanks

PR #14, opened 2018-04-26
make Activity names unique by lesson

Hamms:


Hamms:
@mrjoshida couple notes:

First, this change involves creating a new index on the existing Activities table; given that there are ~4000 entries in that table, I don't expect that to be a huge deal but it's still worth noting.

Second, and more importantly, if there are any existing violations of this new constraint in the table, attempting this migration will cause an error. How worried are you about that happening?

PR #15, opened 2018-04-27
 give Activities a string-based i18n key 

Hamms:
Also optimize string gathering a bit by using `select_related` and only calling `i18n_key` once so we don't waste a bunch of time with fk lookups

### DIFF
--- a/i18n/models.py
+++ b/i18n/models.py
@@ -8,12 +8,13 @@ class Internationalizable:
     @classmethod
     def gather_strings(cls):
         strings = {}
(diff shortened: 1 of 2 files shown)


mrjoshida:
LGTM. I've never seen selecte_related used without arguments, does that just follow all relations?

Hamms:
Yep!

I might want to eventually add the ability for Internationalizable classes to specify the whitelist of relations to grab in that query, but for now it only takes ~90 seconds to run a complete sync so I'm not super worried

PR #16, opened 2018-04-30
Add I18n whitelist logic

Hamms:
Specifically, add two things: the ability for Internationalizable models to have some degree of control over the query used to gather their i18n strings, and the ability for them to declare custom logic to determine whether or not any instance should be translated. (the first thing being used to improve the performance of the second thing)

Internationalizable Pages have some default logic by which they recursively walk through their Page-provided `parent` tree, and only consider themselves to be translatable if all of their parents are.

Curriculum elements, being the root thing we want to actually make i18n decisions based on, define their own logic to only whitelist the `csf-1718` curriculum. This method can easily be overwritten in the future as our requirements change.

Finally, Lessons and Activities also provide their own `should_be_translated` logic. The the former because the default Page implementation is pretty inefficient and Lessons already do their own work to cache the Curriculum relationship, so we might as well use it. The latter because Activities are not Pages and so do not have any default logic.

### DIFF
--- a/curricula/models.py
+++ b/curricula/models.py
@@ -201,6 +201,10 @@ class Curriculum(InternationalizablePage, RichText, CloneableMixin):
     def units(self):
         return Unit.objects.filter(parent=self, login_required=False)
 
+    @property
+    def should_be_translated(self):
+        return self.slug == "csf-1718"
+
     # Hijacking the Mezzanine top menu to control which curricula show on the home page
     @property
     def in_main_menu(self):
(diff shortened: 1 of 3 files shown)


Hamms:
Note that all of the `get_i18n_objects` overrides are completely optional, and are just there to improve performance

Hamms:
Also note that the combination of the extra overhead work and the optimizations result in a net sync time increase of about 45 seconds (from ~90 seconds to ~2min 15 seconds)

PR #21, opened 2018-05-15
Publish curriculum content in all languages

Hamms:
Adds a new manage.py command `publish_i18n`, the companion to `gather_i18n_strings`. This command will run through all translatable content that has been marked as `should_be_translated` and initiate `publish` and `publish_pdfs` operations for that content if those are available.

### DIFF
--- a/bin/sync-i18n.sh
+++ b/bin/sync-i18n.sh
@@ -5,4 +5,4 @@ set -e
 python manage.py gather_i18n_strings
 python manage.py i18n_sync_up
 python manage.py i18n_sync_down
-echo "i18n sync finished, content is ready to publish"
+python manage.py publish_i18n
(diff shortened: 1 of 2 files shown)


Hamms:
@mrjoshida is this actually the right way to programmatically initiate a publish operation? It seems shaky at best.

mrjoshida:
It's the right way because I was trying to write the publish methods in a way that curriculum writers can watch the publish stream when they kick it off. Streaming responses stopped working when I moved to Heroku and I never really got a chance to dig into why. Given that streaming responses haven't been working for quite some time anyway (and I moved that stream into Slack) I'd be open to refactoring the publish methods.

Hamms:
okay, great! I'll leave this as-is for now, and make a note that the publish operation can be simplified at a later date.

Everything else in this change look good, too?

PR #22, opened 2018-05-22
Add i18n_ready field

Hamms:
So units can be manually flagged as ready for internationalization.

Also update the various `should_be_translated` logics to rely on the field. Specifically, curricula are now translated if any of their units have been flagged, and all entities under Unit in the hierarchy (Chapter, Lesson, Activity) are translated iff their Unit is.


Hamms:
Field will automatically show up in the admin panel like so:

image

PR #23, opened 2018-05-23
Maps docs

mrjoshida:
This is the first step in restructing how we organize maps (concept based documentation). This allows maps to be organized into a heirarchy and to reference relevant block documentation. It doesn't yet include any of the template changes needed to actually expose these new connections.


documentation/models.py
```diff
@@ -294,15 +294,26 @@ def get_embed_app(self):
...
-        return '/%s/%s/' % (self.parent.slug, self.slug)
+        url = '/concepts/'
+
+        parent = self.parent
+
+        while parent is not None:
+            url += '%s/' % parent.slug
+            parent = parent.parent
+
+        url += '%s/' % self.slug
```
Hamms: could this just be something like `"/concepts/%s/%s" % (parent.get_absolute_url(), self.slug)`?

documentation/views.py
```diff
@@ -42,16 +44,20 @@ def embed_view(request, slug, ide_slug):
...
-    except Map.DoesNotExist:
-        raise ContinueResolving
+def page_view(request, parents, slug):
+    pages = Map.objects.filter(slug=slug)
+
+    if pages.count() > 1 and parents is not None:
+        # If more than one map was found with the same slug, try narrowing by parent
+        parent_slug = parents.split('/')[-1]
+        if pages.filter(parent__slug=parent_slug).count() > 0:
+            pages = pages.filter(parent__slug=parent_slug)
```
Hamms: nit: I don't remember exactly how Django caches queries, but I think you'll save a select if you do something like

```python
parent_slug = parents.split('/')[-1]
pages_by_parent = pages.filter(parent__slug=parent_slug)
if pages_by_parent.count() > 0:
  pages = pages_by_parent
```

documentation/views.py
```diff
@@ -42,16 +44,20 @@ def embed_view(request, slug, ide_slug):
...
-    try:
-        page = Map.objects.get(slug=slug, parent__slug=curric_slug)
-    except Map.DoesNotExist:
-        raise ContinueResolving
+def page_view(request, parents, slug):
+    pages = Map.objects.filter(slug=slug)
+
+    if pages.count() > 1 and parents is not None:
+        # If more than one map was found with the same slug, try narrowing by parent
+        parent_slug = parents.split('/')[-1]
```
Hamms: any reason in particular we're only using a single parent while also allowing for an arbitrary number in the url? Seems like we should only do one or the other

documentation/views.py
mrjoshida: I think the caching middleware caches view queries, but I think it makes sense to go with your proposal anyway

documentation/views.py
mrjoshida: My thinking was that sibling slugs should be unique, so if we run into a case where more than one map shows up for a given slug filtering by the immediate parent should catch it. I could certainly make it walk up the parent ladder if the duplication isn't resolved by the immediate parent.

documentation/models.py
Hamms: oh, ofc! In that case this looks good

documentation/views.py
Hamms: What do you think about actually enforcing sibling uniqueness with a `unique_together = ("parent", "slug")`?

mrjoshida: In which case we would always query by map and parent slugs here?

Hamms: Yep!

Hamms: Oh dang, I didn't realize Django built derived classes that way.

It's probably not critical to solve this, but it makes me a little uncomfortable to have this potential for unexpected behavior just hanging out.

What do you think about throwing a 500 error in the case where `pages.count() > 1`, with an error message that explains all of this and how to fix it (ie by going into the admin panel and making things appropriately unique)?

documentation/views.py
```diff
@@ -42,16 +44,25 @@ def embed_view(request, slug, ide_slug):
...
-        raise ContinueResolving
+def page_view(request, parents, slug):
+
+    if parents == '':
+        pages = Map.objects.filter(slug=slug)
+    else:
+        parent_slug = parents.split('/')[-1]
+        pages = Map.objects.filter(slug=slug, parent__slug=parent_slug)
+
+    if pages.count() == 0:
```
mrjoshida: @Hamms what do you think of this approach?

Hamms: love it!

PR #24, opened 2018-05-30
Sync translations with Crowdin

Hamms:
Depends on https://github.com/mrjoshida/curriculumbuilder/pull/25

Add two scripts:

- `sync-up` will redact all the i18n source files gathered by `manage.py gather_i18n_sources` and upload the redacted content to crowdin
- `sync-down` will download all (redacted) translations from crowdin, and restore them using the source files.

After both of these scripts have completed successfully, we will probably want to finish by invoking a locale-specific publish operation which will be constructed as a manage.py script. (coming soon in a future PR)

I figure we'll probably want to use something like https://elements.heroku.com/addons/scheduler to run these scripts once a day or so. Thoughts?

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -26,6 +26,7 @@ docker-compose\.dev\.yml
 *.sql
 
 heroku\.env
+.apt
 
 crowdin_credentials.yml
 node_modules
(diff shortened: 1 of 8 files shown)


Hamms:
[create operation to download translations from crowdin to curriculumbuilder](https://trello.com/c/TqJxenWs/134-create-operation-to-download-translations-from-crowdin-to-curriculumbuilder)

PR #29, opened 2018-06-05
set up automated i18n sync on heroku

Hamms:


Hamms:
[automate curriculumbuilder sync (both up and down) on i18n server](https://trello.com/c/vIW19ZnD/136-automate-curriculumbuilder-sync-both-up-and-down-on-i18n-server)

PR #33, opened 2018-06-15
Update a couple links to use get_standards_url rather than assembling it manually

Hamms:
To ensure that they link to the locale-specific version of the standards when in locale mode


Hamms:
Thanks!

PR #40, opened 2018-07-11
Standards opportunities

mrjoshida:
There are some cases where we want to claim a lighter weight standards alignment, which could be better identified as "opportunities for alignment" - specifically when talking about Common Core / NGSS standars in CSF. This adds a separate field to track those standards so that we can call them out as such in lesson plans.

### DIFF
--- a/lessons/admin.py
+++ b/lessons/admin.py
@@ -247,8 +247,8 @@ class LessonForm(ModelForm):
         if self.fields.get('standards', False):
             self.fields['standards'].queryset = standards_queryset
 
-        if self.fields.get('anchor_standards', False):
-            self.fields['anchor_standards'].queryset = standards_queryset
+        if self.fields.get('opportunity_standards', False):
+            self.fields['opportunity_standards'].queryset = standards_queryset
 
         '''
         Optimize loading of blocks with related IDEs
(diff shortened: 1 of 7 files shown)


lessons/admin.py
```diff
@@ -247,8 +247,8 @@ def __init__(self, *args, **kwargs):
         if self.fields.get('standards', False):
             self.fields['standards'].queryset = standards_queryset
 
-        if self.fields.get('anchor_standards', False):
-            self.fields['anchor_standards'].queryset = standards_queryset
+        if self.fields.get('opportunity_standards', False):
+            self.fields['opportunity_standards'].queryset = standards_queryset
```
Hamms: Why are we replacing the anchor standards rather than supplementing them?

mrjoshida: Only in the admin view. Anchor standards have only been used in one course so far (CS in A), so until/unless we decide we want to focus on them again it makes more sense to drop them from the admin view.

Hamms: Does it also make sense to drop them entirely?

mrjoshida: I waffle on that a bit, and we might chose to get rid of them down the road. I'd rather keep the door open as we enter this world of more granularly specifying the extent of alignment - one of the cheif criticisms of our alignment is that we claim to extensive an alignment even within the CSTA standards, so if this "opportunities" message lands we may choose to revisit the idea of specifying a small set of anchors that reflect the core of a lesson in addition to the larger set of all standards this lesson touches on.

PR #42, opened 2018-08-02
enable Resources for translation

Hamms:
Specifically, sync with Crowdin those Resources that are associated with a Lesson that is also being synced, which will cover the 180 Resources that are currently explicitly associated with a CSF lesson.

Also provide an override flag so we can manually enable additional Resources. Say, for example, those Resources that are included merely as markdown links.

image


Hamms:
[Ability to localize links to auxiliary resources on Curriculum Builder (worksheets, pdfs, etc)](https://trello.com/c/ZGTYIQX1/196-ability-to-localize-links-to-auxiliary-resources-on-curriculum-builder-worksheets-pdfs-etc)

PR #45, opened 2018-09-12
Standards export

mrjoshida:
Addressing common request for a downloadable standards spreadsheet.


standards/views.py
```diff
@@ -37,6 +38,26 @@ def by_curriculum(request, slug):
...
+    writer.writerow(['curriculum', 'unit', 'lesson #', 'lesson name',
+                     'standard framework', 'standard', 'cross curricular opportunity'])
+    for unit in curriculum.units:
+        for lesson in unit.lessons:
+            for standard in lesson.standards.all():
+                writer.writerow([curriculum.slug, unit.slug, 'lesson %d' % lesson.number, lesson.title,
+                                 standard.framework.slug, standard.shortcode, ''])
+            for standard in lesson.opportunity_standards.all():
+                writer.writerow([curriculum.slug, unit.slug, 'lesson %d' % lesson.number, lesson.title,
+                                 standard.framework.slug, standard.shortcode, 'X'])
```
Hamms: nit: is there a specific reason to use `''`/`'X'` rather than `true`/`false`?

standards/views.py
```diff
@@ -53,6 +74,26 @@ def by_unit(request, slug, unit_slug):
...
+    writer.writerow(['curriculum', 'unit', 'lesson #', 'lesson name',
+                     'standard framework', 'standard', 'cross curricular opportunity'])
+    for lesson in unit.lessons:
+        for standard in lesson.standards.all():
+            writer.writerow([curriculum.slug, unit.slug, 'lesson %d' % lesson.number, lesson.title,
+                             standard.framework.slug, standard.shortcode, ''])
+        for standard in lesson.opportunity_standards.all():
+            writer.writerow([curriculum.slug, unit.slug, 'lesson %d' % lesson.number, lesson.title,
+                             standard.framework.slug, standard.shortcode, 'X'])
+    return response
```
Hamms: it might make things simpler to give Lesson something like a `to_csv` method that can be used here and above

mrjoshida: This is a min bar to address the immediate need. I plan to come back here and flesh out a better solution later.

standards/views.py
mrjoshida: switched

PR #46, opened 2018-10-02
Restructure URLs for level embedding

mrjoshida:
A while back a change was introduced to levelbuilder that restructured embedded map levels so that they could be accessed from studio.code.org instead of curriculum.code.org. This broke our ability to load those embedded levels in the code studio pullthrough because we are unable to load studio.code.org urls in an iFrame. Ideally I'd like to resolve this such that we aren't restructuring these urls on both sides, but for now this is necessary to fix the code studio pullthrough.


mrjoshida:
/curriculum/csd-18/unit3/6/overview/index.html which needs to be restructured to curriculum.code.org/csd-18/unit3/6/overview/index.html

/docs/applab/setText/index.html which needs to be docs.code.org/applab/setText/index.html

PR #47, opened 2018-10-05
Remove "All Handouts PDF" link from CSF

mrjoshida:
Compiled resource PDFs have been broken in CSF for a while now and we receive consistent tickets about it. Removing the link until I have time to figure out the root cause.


mrjoshida:
It's tracked in my backlog on smartsheet

PR #48, opened 2018-10-06
Restructure maps to live in heierarchy under docs.code.org instead of singletons under a given course

mrjoshida:
Reopening using django page tree logic to organize concept maps


mrjoshida:
@Hamms I took another whack at moving the maps to documentation with a simpler url schema. This assumes a top level map with the slug "concepts" with other maps hung underneath. Eg:

image

Design could still use some love, but I want to get this out so we can fix some maps that have typos using the new url structure

image

curriculumBuilder/settings.py
```diff
@@ -50,11 +50,16 @@
...
+    # Used to generate the main curriculum.code.org course list
     (1, _("Main Curriculum Listing"), "pages/menus/dropdown.html"),
-    (2, _("Unused"), "pages/menus/tree.html"),
-    (3, _("Unused"), "pages/menus/footer.html"),
+    # Used to generate the docs.code.org/concepts/ menu
+    (2, _("Concept Maps"), "pages/menus/tree.html"),
+    #(3, _("Unused"), "pages/menus/footer.html"),
 )
 
+# Don't add new pages to any menuds by default
```
Hamms: nit: s/menuds/menus

documentation/models.py
```diff
@@ -327,12 +325,14 @@ def publish(self, children=False):
...
                 logger.exception('Failed to publish %s' % self)
 
+    '''
     def save(self, *args, **kwargs):
 
         if not self.slug:
             self.slug = slugify(self.title)[:255]
 
         super(Map, self).save(*args, **kwargs)
+    '''
```
Hamms: should this just be removed?

documentation/templates/documentation/map.html
```diff
@@ -0,0 +1,101 @@
...
+{% load staticfiles %}
+{% load mezzanine_tags %}
+{% load keyword_tags %}
+
+{% block meta_keywords %}{% metablock %}
+    {% keywords_for map as keywords %}
+    {% for keyword in keywords %}
+        {% if not forloop.first %}, {% endif %}
+        {{ keyword }}
+    {% endfor %}
```
Hamms: is there a reason this isn't just a `join`?

documentation/urls.py
```diff
@@ -6,10 +6,11 @@
 urlpatterns = [
     multiurl(
         # url(r'^$', views.index, name='home'),
+        url(r'^(?P<slug>.*)/$', views.map_view, name="map_view"),
+        # url(r'^concepts/$', views.maps_view, name='maps_view'),
+        # url(r'^concepts(?P<parents>.*)/(?P<slug>[-\w.]+)/$', views.page_view, name='page_view'),
```
Hamms: should these be removed?

documentation/templates/documentation/map.html
mrjoshida: That's the pattern mezzanine uses everywhere else, the join template tag definitely would be cleaner

documentation/urls.py
mrjoshida: Yup

documentation/models.py
mrjoshida: Yup

PR #50, opened 2018-10-25
Add option to exclude children when cloning.

mrjoshida:
The inital cloning feature assumed that we are always cloning with all children. In reality, it's more reasonable to clone the course empty and then pull in a clone of each unit one at a time as we are prepared to write them. This first step adds a checkbox to exclude children while cloning. Still ToDo:
* Pull out admin pane into a partial template
* Pull out admin js into a separate file
* Add dropdown to clone a unit into a different course

### DIFF
--- a/curricula/templates/curricula/curriculum.html
+++ b/curricula/templates/curricula/curriculum.html
@@ -50,6 +50,10 @@
                 {% else %}
                     <em>Can't publish, check settings in admin</em>
                 {% endif %}
+                <h2>deepSpaceCopy</h2>
+                <button class="btn btn-primary" id="clone_this" data-pk="{{ curriculum.pk }}" data-type="Curriculum">Clone</button><i id="clone_spinner"></i><br/>
+                <input type="checkbox" name="children" id="clone_children"> Clone Children</input>
+                <pre id="clone_results"></pre>
             </div>
         </div>
     </div>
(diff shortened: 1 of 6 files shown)


curricula/templates/curricula/curriculum.html
```diff
@@ -217,6 +221,33 @@ <h2>{{ topic.name }}</h2>
...
+                var pk = $(this).attr('data-pk');
+                var type = $(this).attr('data-type');
+                var children = $('#clone_children').prop('checked');
+                $('#clone_spinner').addClass('fa fa-cog fa-spin');
+
+                $.ajax({
+                     type: "POST",
+                     url: "/clone/",
+                     data: {pk: pk, type: type, children: children},
+                     timeout: 9999999
```
Hamms: this should probably instead either be left blank for an infinite timeout, or given an actual value

curricula/templates/curricula/curriculum.html
```diff
@@ -217,6 +221,33 @@ <h2>{{ topic.name }}</h2>
...
+                        setTimeout(function () {
+                            window.location.href = response.redirect_url;
+                        }, 500);
+                    }
+                }).fail(function (response) {
+                    $("#clone_this").addClass('btn-warning').text("Failed");
+                    $("#clone_results").text(JSON.stringify(response, undefined, 2));
+                    $('#clone_spinner').removeClass('fa fa-cog fa-spin');
+                });
+            });
```
Hamms: looks like this chunk of code gets repeated quite a bit throughout the templates; can it be extracted into a partial?

mrjoshida:
Good call on the children, passing that as True now when cloning with children

curricula/templates/curricula/curriculum.html
mrjoshida: Yup, I've got a todo in unit.html to clean that up. There's a bunch of refactoring I want to do in the templates when I have some time.

curricula/templates/curricula/curriculum.html
mrjoshida: Noted, I'll hit that when I factor this all out

PR #56, opened 2018-12-12
Update README.md with OS X setup instructions

davidsbailey:


davidsbailey:
@mrjoshida where does your virtualenv directory live? I'm seeing it as untracked files within my repo, which is getting annoying. I could see (1) putting it outside of the repo directory (does this work?) or (2) giving it a certain name inside the repo directory and then ignoring it via `.gitignore`.

PR #58, opened 2018-12-18
replace placeholder with unit name when displaying activity.content for code studio lessons

davidsbailey:
screenshot:
screen shot 2018-12-17 at 4 28 37 pm

### DIFF
--- a/curricula/templates/curricula/partials/lesson.html
+++ b/curricula/templates/curricula/partials/lesson.html
@@ -1,5 +1,6 @@
 {% load mezzanine_tags %}
 {% load staticfiles %}
+{% load with_unit %}
 
 <div class="container">
     <div class="header printbackground">
@@ -26,7 +27,7 @@
                 {% for activity in lesson.activity_set.all %}
                     <div class="activity">
                         <h2>{{ activity.name }}{% if activity.time %} ({{ activity.time }}){% endif %}</h2>
-                        {{ activity.content|richtext_filters|safe }}
+                        {{ activity.content|with_unit:unit|richtext_filters|safe }}
                     </div>
                 {% endfor %}
 
--- a/curricula/templates/curricula/codestudiolesson.html
+++ b/curricula/templates/curricula/codestudiolesson.html
@@ -1,6 +1,6 @@
 {% extends "basecurriculum.html" %}
 
-{% load staticfiles mezzanine_tags keyword_tags comment_tags i18n %}
+{% load staticfiles mezzanine_tags keyword_tags comment_tags i18n with_unit %}
 
 {% block meta_title %}{{ lesson.meta_title }}{% endblock %}
 
@@ -192,7 +192,7 @@
         <div class="activity">
             {% editable activity.name activity.time activity.content %}
                 <h2>{{ activity.name }}{% if activity.time %} ({{ activity.time }}){% endif %}</h2>
-                {{ activity.content|richtext_filters|safe }}
+                {{ activity.content|with_unit:unit|richtext_filters|safe }}
             {% endeditable %}
         </div>
     {% endfor %}
--- a/curricula/templatetags/with_unit.py
+++ b/curricula/templatetags/with_unit.py
@@ -1,0 +1,12 @@
+from django import template
+
+register = template.Library()
+
+@register.filter(name='with_unit')
+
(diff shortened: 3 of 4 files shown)


curricula/templates/curricula/partials/lesson.html
```diff
@@ -26,7 +27,7 @@ <h1>Teaching Guide</h1>
                 {% for activity in lesson.activity_set.all %}
                     <div class="activity">
                         <h2>{{ activity.name }}{% if activity.time %} ({{ activity.time }}){% endif %}</h2>
-                        {{ activity.content|richtext_filters|safe }}
+                        {{ activity.content|with_unit:unit|richtext_filters|safe }}
```
davidsbailey: after updating http://localhost:8000/csf-18/coursec/4/ to contain this placeholder, what url could I navigate to in order to exercise this line of code? or should I omit this for now?

davidsbailey:
Josh and Elijah, please take a look! This should be easy to add in more places, anywhere we have access to a unit object that is within `curricula/templates`. However, I want to start this off as a small change initially until we can verify that it's working in prod.

curricula/templates/curricula/partials/lesson.html
mrjoshida: Not sure I understand the question. Am I correct in reading that I'd use this tag like `[link to script](//studio.code.org/s/{{{unit_name}}})`

davidsbailey: Correct, that's how you'd use this new syntax within an activity's content.

The background for my question is, when I use the admin interface to edit an activity's content to contain the `{{{unit_name}}}` placeholder, and then I view it at a url like http://localhost:8000/csf-18/coursec/4/ , I can see that the substitution is working, and I can tell by debugging that the code which is doing the substitution is at   https://github.com/mrjoshida/curriculumbuilder/pull/58/files#diff-a1fb870305086ceb34101808f693a4c2R195

I also made this code change on partials/lesson.html because it appears to render activity.content, but I can't figure out where (if anywhere) the activity content rendered by this partial would show up within my local curriculum builder. If it shows up somewhere, I'd like to verify that the substitution is working.

From looking at the code, it seems like this only ever shows up in unit_view or unit_compiled, which in turn only get used to generate URLs. But I'm new to django, so I just wanted some confirmation that I'm not missing something.

davidsbailey: a math teacher might ask me to show my work 😛 
* partials/lesson.html appears to only be rendered here:
https://github.com/mrjoshida/curriculumbuilder/blob/a1d7f13701c6eeea648b12f1a88edbd8c74986e4/curricula/templates/curricula/unit_lessons.html#L209-L214
* unit_lessons.html appears to only be rendered in https://github.com/mrjoshida/curriculumbuilder/blob/9736e85db46ded446f5b8c948ac73286cf6daaac/curricula/views.py#L115 and https://github.com/mrjoshida/curriculumbuilder/blob/9736e85db46ded446f5b8c948ac73286cf6daaac/curricula/views.py#L404
* `unit_view` and `unit_compiled` appear to only be mentioned as inputs to things that generate urls.

mrjoshida: the the lessons partial is used in the unit_compiled view is used to generate the unit PDFs through eg  http://localhost:8000/csf-18/coursec/compiled/

davidsbailey: Perfect, thank you Josh! That is working:
screen shot 2018-12-19 at 1 31 50 pm

davidsbailey:
Here are the other instances of activity.content that I have not updated yet:
screen shot 2018-12-19 at 1 32 59 pm

I don't think I need to update the ones that have to do with `pl` or `hoc`, but do I need to update pages/lesson.html? again, where would I see that rendered in curriculumbuilder @mrjoshida ?

mrjoshida:
No need to update pages/lesson.html - I need to do a cleanup pass to get rid of templates we're not using anymore.

davidsbailey:
Excellent. Thanks both for the detailed review!

davidsbailey:
oh, here is the screenshot I meant to include originally in the description:
screen shot 2018-12-17 at 3 58 05 pm

PR #63, opened 2019-02-21
Fix documentation url routing

mrjoshida:
At some point recently the documentation routes stopped working. This was likely caused by some of the changes I made to move map levels into docs.code.org. This corrects for that issue as well as fixing a long standing bug that prevented the admin "view on site" button from using the correct local route for ides and blocks

### DIFF
--- a/documentation/admin.py
+++ b/documentation/admin.py
@@ -42,6 +42,10 @@ class IDEAdmin(PageAdmin, VersionAdmin):
         }),
     )
 
+    # Override the "view on site" button to use local /documentation/ instead of docs.code.org path
+    def view_on_site(self, obj):
+        return '/documentation%s' % obj.get_absolute_url()
+
 
 class BlockForm(ModelForm):
     def __init__(self, *args, **kwargs):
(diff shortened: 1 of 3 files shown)


documentation/admin.py
```diff
@@ -95,6 +99,10 @@ class BlockAdmin(PageAdmin, VersionAdmin):
         models.CharField: {'widget': Textarea(attrs={'rows': 2})},
     }
 
+    # Override the "view on site" button to use local /documentation/ instead of docs.code.org path
+    def view_on_site(self, obj):
+        return '/documentation%s' % obj.get_absolute_url()
```
Hamms: why doesn't `get_absolute_url` include this path?

mrjoshida: Because the path that gets used when things are actually published doesn't include documentation. This is because we're trying to publish both docs.code.org and curriculum.code.org from the same same place (docs.code.org on s3 lives at curriculum.code.org/documentation)

Hamms: Oooh, that's unfortunate. Makes sense, though

PR #66, opened 2019-02-27
Update setup

dmcavoy:
Update README with more steps to help guide setup such as the local_settings file and making sure you are a contributor on the project.

### DIFF
--- a/README.md
+++ b/README.md
@@ -65,8 +65,7 @@ heroku pg:pull DATABASE_URL curriculumbuilder -a curriculumbuilder
 8. set up local_settings.py
 
 ```
-create curruculumbuilder/curriculumBuilder/local_settings.py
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -64,11 +62,24 @@ heroku login # need credentials from Josh C.
 heroku pg:pull DATABASE_URL curriculumbuilder -a curriculumbuilder
 ```
 
-8. run the server
+8. set up local_settings.py
+
+```
+create curruculumbuilder/curriculumBuilder/local_settings.py
+copy starter settings from local_settings.py.example
```
davidsbailey: we should already be in the curriculumbuilder directory. so better to just list the exact command:
```
cp curriculumBuilder/local_settings.py.example curriculumBuilder/local_settings.py
```

PR #67, opened 2019-02-28
Assessment Field Lesson Front Page

dmcavoy:
Adds a column in Lessons table for assessment information.


dmcavoy:
I know with Rails migrations we try to do them as a separate PR. Should I do that here as well? Should I include the models file since it actually is what generates the migration?

davidsbailey:
Yes, I think putting the migration in a separate PR which merges first sounds like a great practice for us to follow here too.

dmcavoy:
Alright, I pulled out the changes that were not the models.py update which generates the migration.

PR #73, opened 2019-03-12
Level icons now show the icon instead of a blank square

dmcavoy:
### Before
Screen Shot 2019-03-06 at 1 04 13 PM

### After
Screen Shot 2019-03-12 at 7 39 54 PM

### DIFF
--- a/lessons/static/css/commonlesson.css
+++ b/lessons/static/css/commonlesson.css
@@ -793,11 +793,6 @@ th a:link, th a:visited, th a:hover, th a:active {
 }
 */
 
-.level-icon:before {
-    font: normal normal normal 14px/1 FontAwesome;
-    font-size: inherit;
-}
-
 /* Concept Levels */
 
 /* Text */
@@ -843,7 +838,7 @@ th a:link, th a:visited, th a:hover, th a:active {
 .type-Calc .level-icon:before,
 .level-icon:before
 {
-    content: '\f108';
+    content: "\f108";
 }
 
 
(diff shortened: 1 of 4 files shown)


lessons/static/css/commonlesson.css
```diff
@@ -793,11 +793,6 @@ th a:link, th a:visited, th a:hover, th a:active {
 }
 */
 
-.level-icon:before {
-    font: normal normal normal 14px/1 FontAwesome;
-    font-size: inherit;
-}
```
davidsbailey: its possible these styles were applied by css class `fa` in an older version of font awesome, so when we moved to a newer version, these stopped being the right styles to apply. this is just a guess though 🤷‍♂️ in either case, the approach we're moving toward seems more future-proof than the one you're removing. 👍

islemaster:
Hey Dani! I'm still not seeing these icons on Curriculum Builder ([example](https://curriculum.code.org/csd-18/unit2/4/)).  Is there a final deploy step we're still waiting on?

davidsbailey:
@islemaster It seems that a deploy step is missing. These are the only two places in the code I see `level-icon`, and I am not seeing the `fa` class show up on the missing icons at the link you shared. 

@mrjoshida is there something we need to do to make a deploy happen?

dmcavoy:
@davidsbailey and @islemaster - They show on all codecurricula.com links but in order for them to show on the curriculum.code.org the course content has to be published. I let the curriculum team know that that was the case and they are in charge of publishing their courses when it is appropriate - sometimes they have things in the courses they don't want to publish yet

PR #74, opened 2019-03-13
Add annotated strings from the template files to the translation pipeline

bethanyaconnor:
This adds the ability to annotate text in HTML template files to be translated. This uses Django’s [translation](https://docs.djangoproject.com/en/2.1/topics/i18n/translation/#internationalization-in-template-code) functionality, which gathers all strings into a .po file. We then upload this file to Crowdin (Crowdin supports .po files) and download the files back into the directory associated with the language. After compiling the translations (performed in the down step), Django will replace the strings with translations if they’re available.

Upsides to this approach:
- Very little additional code to be written as we’re leveraging Django functionality
- Fully supported by Crowdin
- Can be expanded easily to include developer added strings not in the templates (I'm currently not sure of the scope of this though)

Downsides to this approach:
- We have to support .po file translations
- The .po file is automatically named “django.po” which isn’t very useful to translators. I’ve got a hacky fix to rename it to miscellaneous.po

This PR includes annotations for one file which was used for testing/prototyping.


i18n/config/crowdin.yml
```diff
@@ -43,6 +43,40 @@ files: [
...
+  # Source files filter
+  # e.g. "/resources/en/*.json"
+  #
+  "source" : "locale/en/**/*.po",
+
+  #
+  # where translations live
+  # e.g. "/resources/%two_letters_code%/%original_file_name%"
+  #
+  "translation" : "/locale/%locale_with_underscore%/LC_MESSAGES/django.po",
```
Hamms: Can we override the [`LOCALE_PATHS`](https://docs.djangoproject.com/en/1.8/ref/settings/#std:setting-LOCALE_PATHS) setting instead of creating a separate `locale` directory?

i18n/management/commands/gather_i18n_strings.py
```diff
@@ -25,3 +26,11 @@ def handle(self, *args, **options):
                 outpath = os.path.abspath(os.path.join(source_dir, model.__name__ + ".json"))
                 with open(outpath, 'w') as outfile:
                     json.dump(strings, outfile, indent=4, sort_keys=True)
+
+        # Gather strings for translation in html files.
+        # For some reason makemessages won't create the locale/ directory but will create
+        # the subdirectories
+        subprocess.call(["mkdir", "locale/"])
+        subprocess.call(["python", "manage.py", "makemessages", "-l", "en"])
```
Hamms: We can use [`django.core.management`](https://docs.djangoproject.com/en/1.8/ref/django-admin/#running-management-commands-from-your-code) rather than `subprocess` for calling the manage.py commands

i18n/management/commands/gather_i18n_strings.py
```diff
@@ -25,3 +26,11 @@ def handle(self, *args, **options):
                 outpath = os.path.abspath(os.path.join(source_dir, model.__name__ + ".json"))
                 with open(outpath, 'w') as outfile:
                     json.dump(strings, outfile, indent=4, sort_keys=True)
+
+        # Gather strings for translation in html files.
+        # For some reason makemessages won't create the locale/ directory but will create
+        # the subdirectories
+        subprocess.call(["mkdir", "locale/"])
```
Hamms: we can use [`os.mkdir`](https://docs.python.org/2/library/os.html#os.mkdir) for this

i18n/management/commands/gather_i18n_strings.py
```diff
@@ -25,3 +26,11 @@ def handle(self, *args, **options):
...
+        subprocess.call(["python", "manage.py", "makemessages", "-l", "en"])
+        # Django.po isn't a very useful file name
+        subprocess.call(["mv", "locale/en/LC_MESSAGES/django.po", "locale/en/LC_MESSAGES/miscellaneous.po"])
```
Hamms: I'm not convinced `miscellaneous` is all that much more useful than `django`. I think I'd prefer to stick with the default here, since it's what django expects

Hamms:
It also looks like `makemessages` right now collects a _ton_ of strings. We might want to audit our existing templates to make sure these are all strings we _actually_ care about translating.

i18n/management/commands/gather_i18n_strings.py
bethanyaconnor: Removed. We can add it back if we think of a better name.

bethanyaconnor:
I think this will work with our existing solution, though I'm not 100% sure. My understanding is that the HTML is published to S3 and then fetched when the page is loaded. This solution would render the translated strings if available, so I *think* the publish operation would work correctly. It's definitely something to look out for though.

FWIW if I run publish_i18n locally (and use a local folder), it does show translated strings in the rendered page. But I'm not sure if that uses the same paths or not.

Hamms:
Hmm, good point. The publish operation is run from the same ephemeral dyno that the sync is run from, so it should indeed work. This would, however, remain an issue for the https://curriculum.code.org/ django site itself.

I'm inclined to say it would be worth looking into a solution that works for both, but not investing too much time into it. If you can't find one, though, we should track finding a solution as a new work item.

i18n/management/commands/gather_i18n_strings.py
bethanyaconnor: Done.

i18n/management/commands/gather_i18n_strings.py
bethanyaconnor: Done.

i18n/config/crowdin.yml
bethanyaconnor: Sort of. I can't find a way to get makemessages to send the gathered strings to a custom directory. I can get Django to find the translations in a translations/ directory, but the format the language codes need to be in is different than the format for the content translations (es_MX vs es-mx). I've made the change to put translations in a translations/ directory, lmk what you think.

We can also do follow-up work to try to standardize the locale format. The underscore format is enforced by the [Django translation system](https://docs.djangoproject.com/en/1.8/topics/i18n/translation/#how-django-discovers-translations) but I'm not sure if the dashed format is required for content translation to work.

curriculumBuilder/settings.py
```diff
@@ -134,6 +134,11 @@
     (LANGUAGE_CODE_DO_TRANSLATION, _('Translate')),
 )
 
+LOCALE_PATHS = (
+    os.path.join(BASE_DIR, 'translations/'),
```
Hamms: I believe this should be `os.path.join(I18nFileWrapper.static_dir(), 'translations')`

bethanyaconnor:
That seems reasonable. I would prefer to support one locale format from Crowdin. The difference in the source folders seems to be an even tradeoff to moving files around in the translation script to me.

I'll make the change to support the underscored locales for the rest of our translations in a separate PR and update the gather_i18n_strings to move the django.po file in this one.

PR #75, opened 2019-03-14
enable syncing Slovak translations

Hamms:


Hamms:
Also created a Jira ticket to track the fact that I had to update two settings to do this: https://codedotorg.atlassian.net/browse/FND-300

PR #81, opened 2019-03-21
Add more sync logging

Hamms:
Follow-up to https://github.com/mrjoshida/curriculumbuilder/pull/80, which made the i18n sync much less verbose than it used to be:

image

This adds a little bit _more_ verbosity, where appropriate.


i18n/management/commands/publish_i18n.py
```diff
@@ -1,56 +1,50 @@
...
-                for language_code, _ in settings.LANGUAGES:
-                    if language_code == settings.LANGUAGE_CODE:
-                        continue
-
+                for language_code in language_codes:
                     translation.activate(language_code)
                     if hasattr(obj, 'publish'):
                         list(obj.publish(silent=True))
                     if hasattr(obj, 'publish_pdfs'):
                         list(obj.publish_pdfs(silent=True))
```
bethanyaconnor: Is it worth logging failures as well?

Hamms: right now, failures in the publish operation are already noisy! A lot of the motivation here was to clear up the _other_ sources of noise, so those errors stand out more.

I would _also_ like to add the ability to track specifically whether or not the operations we're kicking off at various parts of this process (the up/downloads from crowdin, the redaction/restoration, the publish, etc) succeeded or not, and log those explicitly. That's part of what I'm going to be working on later this week.

bethanyaconnor: Sounds good. Thanks for the explanation!

PR #85, opened 2019-03-25
 Add ability to specify whether or not a model should be redacted

Hamms:
[Jira](https://codedotorg.atlassian.net/browse/FND-301)

This is needed specifically for Resources; one of the fields we are
translating for them is a raw URL, which gets parsed as a markdown URL
element by the redaction process, resulting in broken links.

More generally, it's not good for the system to assume that everything
needs to be redacted, or to characterize "redaction" as being synonymous
with "preparing to uplod to crowdin".

This adds the ability for models to specify whether or not they should
be redacted, and adds awareness to the i18n sync process to take
advantage of that. It also renames the directory into which we put
strings to be uploaded to crowdin to make it more clear that redaction
is merely a _part_ of the process.

TODO in a follow-up PR: Make it possible to specify not just a subset of
models to redact, but a subset of fields on a model. (Elijah to add a backlog
item to track this work)


i18n/management/commands/i18n_sync_up.py
```diff
@@ -14,21 +15,28 @@ class Command(BaseCommand):
...
         log("I18n Sync Step 2 of 4: Upload source strings to Crowdin")
         source_dir = os.path.join(I18nFileWrapper.static_dir(), 'source')
-        redacted_dir = os.path.join(I18nFileWrapper.static_dir(), 'redacted')
+        upload_dir = os.path.join(I18nFileWrapper.static_dir(), 'upload')
 
-        if not os.path.exists(redacted_dir):
-            os.makedirs(redacted_dir)
+        # Strings uploaded should include all source strings, some of which
+        # will need to be redacted.
+        copy_tree(source_dir, upload_dir)
```
bethanyaconnor: Why is this necessary?

Hamms: We need the `upload` directory to contain all the stuff that gets uploaded to crowdin. That includes redacted content, so we _also_ need somewhere to keep unredacted ("source") content so we can restore the translations.

Does that answer your question?

bethanyaconnor: Makes sense. Thanks!

PR #86, opened 2019-03-27
add a basic test for Standards CSV functionality

Hamms:
(including new unicode support)


Hamms:
Oh, interesting. Looks like this is failing because the views.py file we're using requires `wkhtmltopdf` to be installed (even though the specific view we're testing does not).

I'll look into adding that to the travis build.

PR #89, opened 2019-03-28
Don't try to publish PDFs for Curriculum

bethanyaconnor:
1. They're broken for English and non-English
2. They don't seem to be used anywhere (and there's no UI element that directs you to them, unlike the courses)

There are still some courses where PDF generation fails (Course D and Express in particular) due to a faulty link in some content. However, the publish command does continue locally, unlike on the Dyno, and I have a theory it's due to a timeout as it would try to publish CSF-1718 next, which is *large*.

~This PR also prevents us from syncing in-tl, which I don't believe we need to be generating.~

### DIFF
--- a/i18n/management/commands/publish_i18n.py
+++ b/i18n/management/commands/publish_i18n.py
@@ -5,7 +5,7 @@ from django.conf import settings
 from django.core.management.base import BaseCommand
 from django.utils import translation
 
-from i18n.management.utils import log, should_publish_pdf, should_sync_model
+from i18n.management.utils import log, should_sync_model
 
 class Command(BaseCommand):
     def should_publish_model(self, model):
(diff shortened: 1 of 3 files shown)


Hamms:
> This PR also prevents us from syncing in-tl, which I don't believe we need to be generating.

We do, actually, so we can host in-context translation from `curriculum.code.org` rather than from `codecurricula.com`

Hamms:
Should we just remove the `publish_pdf` method from the Curriculum model altogether?

bethanyaconnor:
I'm not sure if we should remove that yet. I'm still trying to figure out for sure whether that's used in an internal process, though I couldn't see one.

i18n/management/commands/publish_i18n.py
```diff
@@ -43,13 +42,8 @@ def handle(self, *args, **options):
                     translation.activate(language_code)
                     if hasattr(obj, 'publish'):
                         list(obj.publish(silent=True))
-                    # PDF generation is currently broken in such a way that
-                    # when we attempt to publish a pdf, it not only doesn't
-                    # publish but stops trying to publish anything after this
-                    # step in the sequence.
-                    # TODO: reenable this step once PDF generation is fixed
```
Hamms: :tada:

i18n/management/utils.py
```diff
@@ -7,6 +7,8 @@
 
 from i18n.models import Internationalizable
 
+from curricula.models import Curriculum
+
```
Hamms: nit: we don't need this anymore, either

PR #90, opened 2019-03-28
Speed up "publish" operation

Hamms:
It turns out, most of the time taken up by the publish operation was actually spent repeatedly retrieving all the contents in the S3 bucket into a "preloaded metadata" memory, then ignoring that content once it was retrieved. (see https://stackoverflow.com/a/21121924/1810460 for more details)

To address this, we provide an implementation of S3BotoStorage for Jackfrost that doesn't use preloading at all, which should make publish operations _much_ faster.

Also, provide a local memory cache (rather than no cache at all) for instances that don't provide redis (such as local developer instances)


Hamms:
Also tagging @joshlory if you curious

Hamms:
I realized after looking at the docs that there is actually a much simpler way to disable preloading; PTAL

PR #91, opened 2019-03-29
Removing translation tags from admin templates, plus comment.html

bethanyaconnor:


templates/generic/includes/comment.html
```diff
@@ -19,18 +19,18 @@ <h4 class="media-heading"><a href="{{ comment.url }}">{{ comment.user_name }}</a
         <h4 class="media-heading">{{ comment.user_name }}
     {% endif %}
             <small class="timespan">
-                {% blocktrans with sometime=comment.submit_date|timesince %}{{ sometime }} ago{% endblocktrans %}
+                {% with sometime=comment.submit_date|timesince %}{{ sometime }} ago{% endwith %}
```
Hamms: I'm not super familiar with how the `with` block works; have you verified that this template still renders correctly?

bethanyaconnor: I honestly cannot figure out where this template is used and am tempted to delete it. I think it was used once, so some of that framework remains, like in [Lessons](https://github.com/mrjoshida/curriculumbuilder/blob/576bbf2e6b0377abada80755759a6fc2c9ee4490/lessons/models.py#L286)

Hamms: given its location in `generic`, I would guess it's just something included by default for django as an example of something you can do.

Particularly given [its commit history](https://github.com/mrjoshida/curriculumbuilder/commits/b487fdce5a5f32288e735526eefc57095a8b556d/templates/generic/includes/comment.html)

bethanyaconnor: Agreed. I'm going to delete it -- it's not included in any other template.

PR #99, opened 2019-04-01
Font Awesome Version 4.6.3 and View on Code Studio Icon

dmcavoy:
# Font Awesome Version
Rolls us back to Font Awesome Version 4.6.3 to match Code Studio. (Reverses the work of this commit: https://github.com/mrjoshida/curriculumbuilder/commit/7023033ac29a5950a0e0ff1fc1ff9c9baa0129a8)

# View on Code Studio Icon
By rolling back to the earlier version of Font Awesome we can use the external-link icon again. This also sets it up with the styling we needed to do so it would not show as a blank square.

Screen Shot 2019-04-02 at 10 07 53 AM

### DIFF
--- a/curricula/templates/curricula/partials/code_studio.html
+++ b/curricula/templates/curricula/partials/code_studio.html
@@ -47,6 +47,7 @@
 
                         <a href="{{ level.path|level_link }}" target="_blank" class="level-link">
                             View on Code Studio
+                            <span class="level-link-icon fa"></span>
                         </a>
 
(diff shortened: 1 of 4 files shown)


davidsbailey:
Code changes here look good and seems fine to merge.

per comments in email, this might be fixable by downgrading font awesome from v5.x to 4.x.  before we move on from this, let's make a decision on whether to try harder to get fa-external-link to work.

PR #100, opened 2019-04-01
Fix unit resources pdf to correctly render unicode characters.

bethanyaconnor:
Before: 
Screenshot 2019-04-01 at 3 13 31 PM

After:
Screenshot 2019-04-01 at 3 13 37 PM

### DIFF
--- a/curricula/templates/curricula/lesson_title.html
+++ b/curricula/templates/curricula/lesson_title.html
@@ -1,4 +1,5 @@
 <head>
+    <meta charset="utf-8">
     <style>
         h1, h2, h3 {
             text-align: center;


bethanyaconnor:
It wouldn't surprise me if this fix will be needed again but anything that extends [basecurriculum.html](https://github.com/mrjoshida/curriculumbuilder/blob/720d75ea6cf1d2469d31fa34329fc86988756d4c/templates/basecurriculum.html#L13) already has this line (which includes the compiled unit view). I haven't seen anywhere else that isn't displaying so I'm pretty happy leaving this to be a case-by-case fix.

PR #103, opened 2019-04-05
make the various standards models translatable

Hamms:
[Jira](https://codedotorg.atlassian.net/browse/FND-264)

### DIFF
--- a/standards/models.py
+++ b/standards/models.py
@@ -1,13 +1,15 @@
 from django.db import models
 from django.db.models import Count, Q
 
+from i18n.models import Internationalizable
+
 """
 Top level of a standards framework
 
 """
 
 
-class Framework(models.Model):
+class Framework(Internationalizable):
     name = models.CharField(max_length=255)
     slug = models.SlugField()
     description = models.TextField()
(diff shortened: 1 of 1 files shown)


Hamms:
@mrjoshida can you comment on whether we need _both_ `name` and `description` to be translated for these models? Looking at the data, it appears a lot of these models use the same values for both fields, but I can't find anything in the code enforcing that. Do we actually use both fields?

mrjoshida:
Long term we have a goal of making the name shorter and more consumable than the full length. Currently there are a handful of standards that overflow the character limit for name, but that's the only real difference. A higher level question is whether these are valuable to translate in the first place, given that these are local standards and may not be useful outside of the US. For international folks that lesson objectives are maybe more meaningful

Hamms:
Great, thanks!

Regardless of the specific value to non-US users, we want to make it _possible_ to translate these values in pursuit of the ideal that anything that's visible to a user should be able to be viewed in their language.

mrjoshida:
I guess what I'm proposing is that we don't show standards alignment in translated views because they aren't meaningfully international and it feels like a whole lot of text to translate without a ton of value

Hamms:
Note that translated does not strictly imply non-US

PR #104, opened 2019-04-09
Batch LevelBuilder sync

mrjoshida:
This adds the code studio sync button to both unit and curriculum levels so we can run batch updates. While I was in there I refactored out the admin pane to a separate template and a lot of the template javascript into individual files.


curricula/templates/curricula/unit.html
```diff
@@ -309,202 +278,6 @@ <h4>
 {% block footer_js %}
 
 {% if user.is_staff %}
-    {% comment %} ToDo: pull this out into a shared js file {% endcomment %}
```
Hamms: :tada:

lessons/models.py
```diff
@@ -423,15 +423,20 @@ def get_levels(self):
...
-            self.stage = data
-            self.save()
-        except Exception:
-            logger.warning("Couldn't get stage details for %s" % self)
+        if not hasattr(self.unit, 'stage_name'):
+            return {'error': 'No stage name for unit', 'status': 404}
+        else:
+            try:
+                url = "https://levelbuilder-studio.code.org/s/%s/stage/%d/summary_for_lesson_plans" % (
+                self.unit.stage_name, self.number)
```
Hamms: nit:
```suggestion
                    self.unit.stage_name, self.number)
```

curricula/templates/curricula/partials/admin_menu.html
```diff
@@ -8,45 +8,62 @@
...
         <div class="drawer-body">
+
             <h2>Add to Log</h2>
             <form method="POST" action="." class="post-form">{% csrf_token %}
                 {{ form.as_p }}
                 <button type="submit" class="save btn btn-primary">Save</button>
             </form>
-            <h3><a href="/page_history/{{ lesson.pk }}/">View history</a></h3>
+            {% if unit %}
+            <h3><a href="unit_feedback/">View unit feedback</a></h3>
```
Hamms: nit: can we try to use the `{% url ... %}` tag as much as possible for things like this?

curricula/templates/curricula/unit.html
```diff
@@ -278,202 +278,6 @@ <h4>
...
-                    }
-                }).fail(function (response) {
-                    $("#clone_this").addClass('btn-warning').text("Failed");
-                    $("#clone_results").text(JSON.stringify(response, undefined, 2));
-                    $('#clone_spinner').removeClass('fa fa-cog fa-spin');
-                });
-            });
-        });
-    </script>
+    <script src="{% static "js/admin_menu.js" %}"></script>
```
Hamms: is there a reason why we only sometimes gate this with a `user.is_staff` check? Seems like that should always happen

mrjoshida: This is gated in `user.is_staff`, it's just hiding above that huge mess of removed lines :)

curricula/templates/curricula/partials/admin_menu.html
mrjoshida: updated

PR #109, opened 2019-04-18
Add gettext to the heroku dynos

Hamms:
Used by the Django translation system, added in
https://github.com/mrjoshida/curriculumbuilder/pull/74


Hamms:
FYI @bethanyaconnor

PR #111, opened 2019-04-18
use locale_name for sync out

Hamms:
In https://github.com/mrjoshida/curriculumbuilder/pull/74, we updated the locations of the translations files for the sync in and down, but needed to also update the other parts of the sync out (ie, restoration and uploading).

In part, that was because we were manually specifying the location of the i18n files in several locations; I've updated all these to use the new I18nFileWrapper.locale_dir method, which should mean that future changes of this nature only have to happen in one place.

I also removed the `languages_mapping` in the crowdin config, since it appears that the `locale_with_underscore` settings means that crowdin does, in fact, use the four-character version

### DIFF
--- a/i18n/config/crowdin.yml
+++ b/i18n/config/crowdin.yml
@@ -36,15 +36,5 @@ files: [
   # e.g. "update_as_unapproved" or "update_without_changes"
   #
   "update_option" : "update_without_changes",
-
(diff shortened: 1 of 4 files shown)


Hamms:
(intentionally merging before review to unblock the i18n sync)

PR #112, opened 2019-04-19
Fix capitalization of in-context-translation pseudo-language

Hamms:
For some reason, when downloading translations using
locale_with_underscore, crowdin correctly uppercases the locale for all
languages EXCEPT the pseudo-language we use to enable in-context
translations. So we have to manually coerce it to the right format.


Hamms:
(merging before review to unblock i18n sync)

PR #117, opened 2019-05-01
Add template translations for lesson details

joshlory:
https://codedotorg.atlassian.net/browse/FND-233
https://codedotorg.atlassian.net/browse/FND-234
https://codedotorg.atlassian.net/browse/FND-235

### DIFF
--- a/curricula/templates/curricula/partials/lesson_pills.html
+++ b/curricula/templates/curricula/partials/lesson_pills.html
@@ -1,18 +1,19 @@
+{% load i18n %}
 {% load mezzanine_tags %}
 <div class="resource-nav btn-toolbar" role="toolbar" aria-label="Lesson Resources">
     <div class="btn-group btn-group-xs" role="group" aria-label="PDF Buttons">
-        <a href="{{ unit.get_absolute_url }}" class="btn" role="button">Unit Overview</a>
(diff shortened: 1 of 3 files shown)


curricula/templates/curricula/partials/lesson_pills.html
```diff
@@ -1,18 +1,19 @@
...
-        <a href="{{ unit.get_resources_pdf_url }}" class="btn" role="button">All Handouts PDF</a>
+        <a href="{{ unit.get_resources_pdf_url }}" class="btn" role="button">{% trans 'All Handouts PDF' %}</a>
         {% endif %}
         {% if changelog.count > 0 %}
 
             <div class="btn-group" role="group">
             <button type="button" class="btn dropdown-toggle" data-toggle="dropdown" aria-haspopup="true"
                     aria-expanded="false">
-                Last updated {{ changelog.first.revision.date_created|date:"n/j/y" }}
+                {% trans 'Last updated' %} {{ changelog.first.revision.date_created|date:"n/j/y" }}
```
bethanyaconnor: "Last updated" looks like it's used in 3 place: unit_pills.html, curriculum_pills.html, and here. I think this PR is good as is, but just noting that for the Jira item.

joshlory:
Good call, fixed in 69b4cb9c6e69d0ee9d4239e0654c13786c3425a5!

PR #118, opened 2019-05-01
Keywords i18n serializer

joshlory:
Allow iterable fields to be added to `internationalizable_fields` (provided a matching serializer). Sample source/Lesson.json:

```js
"coursee/introduction-course-warm-up": {
    "cs_content": "We recognize that every classroom has a spectrum of understanding for every subject. Some students in your class may be computer wizards, while others haven't had much experience at all. In order to create an equal playing (and learning) field, we have developed this \"Ramp-Up Stage\" for Course E. This can be used as either an introduction or a review to the Code.org interface and basic computer science concepts. This stage, along with the three that follow, cover all prerequisites needed to start Course E.", 
    "description": "This lesson will give you practice in the skills you will need for this course.", 
    "keywords": [
        "Sequencing", 
        "Debugging", 
        "Loop", 
        "Ice Age", 
        "Maze", 
        "Artist"
    ], 
    "overview": "In this progression, students will begin with an introduction (or review depending on the experience of your class) of Code.org's online workspace. Students will learn the basic functionality of the interface including the `Run`, `Reset`, and `Step` buttons. Dragging, deleting, and connecting Blockly blocks is also introduced in the beginning video. In the puzzles, students will practice their sequencing and debugging skills in Maze and Artist.", 
    "prep": " - Play through puzzles on [r csf-course-e] to find any potential problem areas for your class.\r\n - Make sure every student has a [r think-spot-journal].", 
    "title": "Introduction to Online Puzzles"
}, 
```

https://codedotorg.atlassian.net/browse/FND-232

### DIFF
--- a/curricula/serializers.py
+++ b/curricula/serializers.py
@@ -42,11 +42,12 @@ class LessonSerializer(serializers.ModelSerializer):
     student_resources = serializers.SerializerMethodField()
     vocab = serializers.SerializerMethodField()
     blocks = serializers.SerializerMethodField()
+    keywords = serializers.SerializerMethodField()
 
     class Meta:
         model = Lesson
         fields = ('title', 'number', 'student_desc', 'teacher_desc',
-                  'student_resources', 'teacher_resources', 'vocab', 'blocks')
+                  'student_resources', 'teacher_resources', 'vocab', 'blocks', 'keywords')
 
     def get_teacher_desc(self, obj):
         return obj.overview
@@ -73,6 +74,9 @@ class LessonSerializer(serializers.ModelSerializer):
         blocks = obj.blocks.all()
         serializer = BlockSerializer(blocks, many=True)
         return serializer.data
+
+    def get_keywords(self, obj):
+        return [str(k) for k in obj.keywords.all()]
 
 
 class UnitSerializer(serializers.ModelSerializer):
--- a/i18n/models.py
+++ b/i18n/models.py
@@ -46,9 +46,16 @@ class Internationalizable(models.Model):
         return cls.objects
 
     @classmethod
+    def get_serializer(cls):
+        name = cls.__name__ + 'Serializer'
+        serializers = __import__('curricula.serializers', fromlist=[name])
+        return getattr(serializers, name, None)
+
+    @classmethod
     def gather_strings(cls):
         strings = {}
         objects = cls.get_i18n_objects()
+        Serializer = cls.get_serializer()
         total = objects.count()
         index = 0
         elapsed = 0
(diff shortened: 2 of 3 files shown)


i18n/models.py
```diff
@@ -62,10 +69,13 @@ def gather_strings(cls):
             if obj.should_be_translated:
                 key = obj.i18n_key
                 strings[key] = {}
+                serialized_obj = {}
+                if Serializer:
+                    serialized_obj = Serializer(obj).data
                 for field in cls.internationalizable_fields():
-                    string = getattr(obj, field)
+                    string = serialized_obj[field] if serialized_obj.has_key(field) else getattr(obj, field)
```
joshlory: I wonder if this gets cleaner by creating serializers for _all_ models, and returning the serialized dictionary filtered by `internationalizable_fields`?

joshlory:
The docs I used while making this PR:

https://www.django-rest-framework.org/api-guide/serializers/
https://www.django-rest-framework.org/api-guide/relations/#generic-relationships
https://stackoverflow.com/questions/547829/how-to-dynamically-load-a-python-class

Another option I didn't pursue:

https://stackoverflow.com/questions/49282422/how-to-find-serializer-for-a-django-model

i18n/models.py
```diff
@@ -45,10 +45,17 @@ def internationalizable_fields(cls):
     def get_i18n_objects(cls):
         return cls.objects
 
+    @classmethod
+    def get_serializer(cls):
+        name = cls.__name__ + 'Serializer'
+        serializers = __import__('curricula.serializers', fromlist=[name])
```
Hamms: could you add a comment here linking to the SO answer that led you to this?

i18n/models.py
Hamms: yeah, I could definitely see that!

Rebrand "gather_strings" to "serialize i18n data" (which is a more accurate and descriptive name, anyway) and do something like require you to define serializers rather than `internationalizable_fields`, or perhaps something that can easily generate serializers for the simple use cases but would let you manually define them for things like this that are doing something a little more complicated.

If I had remembered that django serializers were a thing, I probably would have built it that way to begin with.

curricula/serializers.py
```diff
@@ -42,11 +42,12 @@ class LessonSerializer(serializers.ModelSerializer):
     student_resources = serializers.SerializerMethodField()
     vocab = serializers.SerializerMethodField()
     blocks = serializers.SerializerMethodField()
+    keywords = serializers.SerializerMethodField()
 
     class Meta:
         model = Lesson
         fields = ('title', 'number', 'student_desc', 'teacher_desc',
-                  'student_resources', 'teacher_resources', 'vocab', 'blocks')
+                  'student_resources', 'teacher_resources', 'vocab', 'blocks', 'keywords')
```
bethanyaconnor: I'm not sure what this file does currently but is there any potential risk to adding another field?

joshlory: I believe it'll start showing up on the API view: https://www.codecurricula.com/api/v1/curriculum/csf-1718/coursea/2/

@mrjoshida are these views used? Any downside to adding a few more attributes to the API view?

i18n/models.py
joshlory: Added in commit b154e2c9f029c8c223ed68968d6a5edf62107c38.

joshlory:
@Hamms and I dug into this a bit more ([slack link](https://codedotorg.slack.com/archives/CFTFD6BPV/p1558715561000400)) but didn't find a clean way to override the Mezzanine Keywords model. We're going to move forward with this PR for now, trusting Crowdin's duplicates feature to take care of existing translations if/when we switch to storing keywords all in one place.

curricula/serializers.py
mrjoshida: I can't see any reason that'd be a problem

PR #122, opened 2019-05-08
Assessment Field

mrjoshida:
To better communicate how each course approaches assessment in general, and in particular in the context of aligned standards frameworks, I'm adding an "assessment_commentary" field to both the curriculum and unit models. This field gets rendered at the top of the curriculum and unit standards overview pages.

### DIFF
--- a/curricula/models.py
+++ b/curricula/models.py
@@ -64,6 +64,10 @@ class Curriculum(InternationalizablePage, RichText, CloneableMixin):
     class Meta:
         verbose_name_plural = "curricula"
 
+    @classmethod
+    def internationalizable_fields(cls):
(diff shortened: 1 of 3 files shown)


curricula/models.py
```diff
@@ -62,6 +64,10 @@ class Curriculum(InternationalizablePage, RichText, CloneableMixin):
     class Meta:
         verbose_name_plural = "curricula"
 
+    @classmethod
+    def internationalizable_fields(cls):
+        return super(Curriculum, cls).internationalizable_fields() + ['assessment_commentary']
+
```
mrjoshida: @Hamms is this the correct way to make this field internationalizable?

Hamms: Yep!

PR #123, opened 2019-05-14
Tracking Pixel 

Erin007:
[LP-187](https://codedotorg.atlassian.net/browse/LP-187)

We'd like to start gathering metrics about the usage of unplugged lessons, especially to do a comparison of their usage between teachers who have gone through professional development and teachers who have not. Views of unplugged lesson plans on curriculum.code.org is being used as a proxy for engagement with unplugged lessons.

This PR sets up the templatetag to view a tracking pixel served from studio.code.org on curriculum.code.org pages. The tracking pixel is then used to gather the url for the curriculum page viewed and the current user id, if there is one, and log that information to Redshift. 

[Related studio.code.org PR](https://github.com/code-dot-org/code-dot-org/pull/28532)


curricula/templatetags/tracking_pixel_url.py
```diff
@@ -0,0 +1,15 @@
...
+register = template.Library()
+
+@register.filter(name='tracking_pixel_url')
+
+def tracking_pixel_url(path):
+    # Append a random number so that the tracking_pixel
+    # isn't cached - we want each visit to a url to go
+    # fetch a new tracking pixel image.
+    random_num = random.randint(1,1001)
+    tracking_pixel_url = 'https://studio.code.org/curriculum_tracking_pixel?from=' + urllib.quote_plus(path) + '&r=' + str(random_num)
```
breville: Is the random number required for cache-busting if we set the right cache controls on the server?

templates/basecurriculum.html
```diff
@@ -82,6 +82,7 @@
     <div class="header printbackground">
         <div class="logo">
             <img src="{% static "images/logo.png" %}">
+            <img src='{{ request.path | tracking_pixel_url}}'>
```
breville: What does this do exactly?

Erin007: It's Django shenanigans. 

In the other file, we wrote a custom template filter,
 `tracking_pixel_url`, which takes one parameter, `path`.  This line is how we call that function with `request.path` as an argument. The return value of which, the tracking pixel url that hits studio.code.org, is then formatted as a string and used as the source for the image.

https://docs.djangoproject.com/en/2.2/howto/custom-template-tags/

PR #124, opened 2019-05-19
Make `Vocab` a translatable model

joshlory:
Sample from source/Vocab.json:

```js
"command": {
    "detailDef": "An instruction for the computer. Many commands put together make up algorithms and computer programs. ", 
    "simpleDef": "An instruction for the computer. Many commands put together make up algorithms and computer programs. ", 
    "word": "Command"
}, 
"condition": {
    "detailDef": "A statement that a program checks to see if it is true or false.  If true, an action is taken. Otherwise, the action is ignored.", 
    "simpleDef": "Something a program checks to see if it is true before allowing an action.", 
    "word": "Condition"
}, 
"conditionals": {
    "detailDef": "Statements that only run under certain conditions.", 
    "simpleDef": "Statements that only run under certain conditions.", 
    "word": "Conditionals"
}, 
```

Most of the short/long definitions are the same, but not all.

### DIFF
--- a/lessons/models.py
+++ b/lessons/models.py
@@ -53,7 +53,7 @@ Vocabulary
 
 
 class Vocab(Internationalizable):
-    word = models.CharField(unique=True, max_length=255)
+    word = models.CharField(max_length=255)
     simpleDef = models.TextField()
     detailDef = models.TextField(blank=True, null=True)
     mathy = models.BooleanField(default=False)
@@ -61,6 +61,7 @@ class Vocab(Internationalizable):
     class Meta:
         ordering = ["word"]
         verbose_name_plural = "vocab words"
+        unique_together = ('word', 'mathy')
 
     @property
     def i18n_key(self):
--- a/lessons/models.py
+++ b/lessons/models.py
@@ -53,7 +53,7 @@ Vocabulary
 
 
 class Vocab(Internationalizable):
-    word = models.CharField(max_length=255)
+    word = models.CharField(unique=True, max_length=255)
     simpleDef = models.TextField()
     detailDef = models.TextField(blank=True, null=True)
     mathy = models.BooleanField(default=False)
--- a/lessons/models.py
+++ b/lessons/models.py
@@ -70,10 +70,6 @@ class Vocab(Internationalizable):
     def internationalizable_fields(cls):
         return ['word', 'simpleDef', 'detailDef']
 
-    @classmethod
-    def get_i18n_objects(cls):
-        return super(Vocab, cls).get_i18n_objects()
-
     @property
     def should_be_translated(self):
         return any(lesson.should_be_translated for lesson in self.lesson_set.all())
--- a/lessons/models.py
+++ b/lessons/models.py
@@ -64,7 +64,7 @@ class Vocab(Internationalizable):
 
     @property
     def i18n_key(self):
-        return self.word
+        return slugify(self.get_untranslated_field('word'))
 
     @classmethod
     def internationalizable_fields(cls):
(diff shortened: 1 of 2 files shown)


joshlory:
After seeing how easy this was, should I abandon PR https://github.com/mrjoshida/curriculumbuilder/pull/118 and do the same for Keywords?

It's a built-in mezzanine generic field, but I'd imagine I can monkey-patch it, or create a simple wrapper that also inherits from `Internationalizable`.

lessons/models.py
```diff
@@ -62,6 +62,22 @@ class Meta:
...
+    def i18n_key(self):
+        return slugify(self.get_untranslated_field('word'))
+
+    @classmethod
+    def internationalizable_fields(cls):
+        return ['word', 'simpleDef', 'detailDef']
+
+    @classmethod
+    def get_i18n_objects(cls):
+        return super(Vocab, cls).get_i18n_objects()
```
Hamms: note that it's only necessary to override this if you want to modify it, and the primary reason to modify it is to prefetch related models to make the `should_be_translated` fetches faster. For example:

https://github.com/mrjoshida/curriculumbuilder/blob/b04cec0692d213e7d32d9a7d18b16658c5aeca36/lessons/models.py#L112-L118

lessons/models.py
```diff
@@ -62,6 +62,22 @@ class Meta:
         ordering = ["word"]
         verbose_name_plural = "vocab words"
 
+    @property
+    def i18n_key(self):
+        return slugify(self.get_untranslated_field('word'))
```
Hamms: is this guaranteed to be unique?

bethanyaconnor:
I'd probably prefer creating some sort of `InternationalizableKeyword` model that inherits from `Internationalizable` and `Keyword` (which I'm guessing is along the lines of what you were thinking?), mostly for consistency. It's nice to have all fields follow the same model for i18n.

lessons/models.py
joshlory: Should I add a prefetch here, or ok to remove this entire class method?

lessons/models.py
joshlory: Good call — added a unique constraint in 43e284886e9bd7185d4729419ed1ca05848055bf. But I need to work with @mrjoshida on getting the existing data to not have duplicates:

```py
In [32]: Vocab.objects.filter(word="Algorithm")
Out[32]: [
  <Vocab: Algorithm: A list of steps that you can follow to finish a task.>,
  <Vocab: Algorithm: A list of steps to finish a task. A set of instructions that can be performed with or without a computer. For example, the collection of steps to make a peanut butter and jelly sandwich is an algorithm.>,
  <Vocab: Algorithm: A list of steps to finish a task. >
]
```

joshlory: Ok, we've removed all duplicates! I pulled down the DB locally and verified the new migration succeeds with the most recent data.

joshlory:
@Hamms PTAL!

lessons/models.py
```diff
@@ -62,6 +62,18 @@ class Meta:
...
+
+    @classmethod
+    def internationalizable_fields(cls):
+        return ['word', 'simpleDef', 'detailDef']
+
+    @property
+    def should_be_translated(self):
+        return any(lesson.should_be_translated for lesson in self.lesson_set.all()) and not self.mathy
```
Hamms: nit:
```suggestion
        return not self.mathy and any(lesson.should_be_translated for lesson in self.lesson_set.all())
```

saves us from an expensive query for all mathy values

PR #125, opened 2019-05-20
Assessment and Mini Rubric: Release to All Courses

dmcavoy:
Remove 2019 restriction on assessment design to match removing the flag on code studio.

This impacts the Code Studio Pull-Through. We will now show the assessment marker on any level marked as an assessment in any course. We will also show the mini rubric for any level that has a mini rubric.

Screen Shot 2019-05-20 at 12 18 49 PM

### DIFF
--- a/curricula/templates/curricula/partials/code_studio.html
+++ b/curricula/templates/curricula/partials/code_studio.html
@@ -16,13 +16,9 @@
                         <li class="chunk-header type-{{ level.type }} {{ unit.slug }}-stage-{{ lesson.number }}-level-{{ level.position }}-tab" data-level="{{ level.position }}" data-named="{{ level.named_level }}">
                             {{ level.display_name|default:level.name }}
                             <span class="level-icon fa"></span>
-                            <!-- TODO: Remove outer if statement here when -->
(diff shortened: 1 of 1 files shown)


curricula/templates/curricula/partials/code_studio.html
```diff
@@ -16,13 +16,9 @@
                         <li class="chunk-header type-{{ level.type }} {{ unit.slug }}-stage-{{ lesson.number }}-level-{{ level.position }}-tab" data-level="{{ level.position }}" data-named="{{ level.named_level }}">
                             {{ level.display_name|default:level.name }}
                             <span class="level-icon fa"></span>
-                            <!-- TODO: Remove outer if statement here when -->
-                                    <!-- we want to roll out for all courses -->
-                                    {% if '2019' in level.path %}
```
davidsbailey: this is safe to remove because we've removed the assessment flag from puzzles in all CSF scripts we care about, right?

dmcavoy: Correct - we did the work on code studio to make sure only things we want marked as assessments are marked as assessments so when we pull the information over it should be correct.

PR #126, opened 2019-05-20
hide instructions content via .no-pullthrough

davidsbailey:
partially addresses [LP-180](https://codedotorg.atlassian.net/browse/LP-180)

Specifically, makes it so that instructions content can be excluded from code studio pull-through into curriculumbuilder via the `.no-pullthrough` css class.


davidsbailey:
### before

Screen Shot 2019-05-21 at 10 26 44 AM

### after

Screen Shot 2019-05-21 at 10 26 32 AM

PR #133, opened 2019-06-20
Hide pdf links for lessons and handouts for csf-2019

Erin007:
Lesson plan and handout PDF links for CSF-19 currently dead end on a "Requested Page Not Found" page because there is a problem with PDF generation. 
Screen Shot 2019-06-20 at 5 23 21 PM

The Learning Platform team is tracking a deeper investigation and longer term fix for PDF generation issues ([LP-545 ](https://codedotorg.atlassian.net/browse/LP-545)). 

In the meantime, I've removed the Lesson PDF and Handouts PDF links for Express & Pre Express 2019. 

BEFORE:
links-before

AFTER:
links-after

### DIFF
--- a/curricula/templates/curricula/partials/lesson_pills.html
+++ b/curricula/templates/curricula/partials/lesson_pills.html
@@ -3,8 +3,8 @@
 <div class="resource-nav btn-toolbar" role="toolbar" aria-label="Lesson Resources">
     <div class="btn-group btn-group-xs" role="group" aria-label="PDF Buttons">
         <a href="{{ unit.get_absolute_url }}" class="btn" role="button">{% trans 'Unit Overview' %}</a>
-        {# Removing lessons PDF link from CSF until we can fix it #}
-        {% if unit.has_resource_pdf %}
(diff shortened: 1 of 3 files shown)


curricula/templates/curricula/partials/lesson_pills.html
```diff
@@ -3,7 +3,10 @@
 <div class="resource-nav btn-toolbar" role="toolbar" aria-label="Lesson Resources">
     <div class="btn-group btn-group-xs" role="group" aria-label="PDF Buttons">
         <a href="{{ unit.get_absolute_url }}" class="btn" role="button">{% trans 'Unit Overview' %}</a>
+        {# Removing lessons PDF link from CSF until we can fix it #}
+        {% if unit.has_resource_pdf %}
```
mrjoshida: Worth noting that this is likely to cause us tickets in A-F where those PDFs do work. Is it worth hard coding this to just target express/pre-express until we can fix?

Erin007:
Updated so that Lesson PDF link is hidden for Express and Pre-Express 2019 and Handout link is hidden for all of CSF-2019.

PR #137, opened 2019-06-26
Use strings for i18n keys

Hamms:
Since our i18n values are communicated through JSON, we always want to
use strings. Using non-string values will work fine for source string
generation, since the keys will be converted to strings in the
serialization process; but it will fail to find the translations since
the string keys won't be converted back to integers in the
deserialization process.

### DIFF
--- a/i18n/utils.py
+++ b/i18n/utils.py
@@ -77,7 +77,9 @@ class I18nFileWrapper:
     def get_translated_field(cls, name, i18n_key, field, lang):
         translations = cls._load_translations(name, lang)
         try:
-            return translations[i18n_key][field]
+            # Always use keys to access the translations dict, since it's
+            # loaded from JSON
+            return translations[str(i18n_key)][str(field)]
         except KeyError:
             # Could not find the specified string in the translation file,
             # possibly just because the string is new and has not yet been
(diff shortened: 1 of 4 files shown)


Hamms:
Hmm, the new test passes locally but it looks like Travis isn't able to create a table for the test model like my local environment can; not sure why.

joshlory:
Note for myself: this should fix models like Category, Framework, GradeBand, Objective and Standard not displaying for in-context translation.

image

i18n/utils.py
```diff
@@ -77,7 +77,9 @@ def _load_translations(cls, name, lang):
     def get_translated_field(cls, name, i18n_key, field, lang):
         translations = cls._load_translations(name, lang)
         try:
-            return translations[i18n_key][field]
+            # Always use keys to access the translations dict, since it's
+            # loaded from JSON
+            return translations[str(i18n_key)][str(field)]
```
joshlory: Does this make the `str(self.pk)` change above not needed?

Hamms: Yep! But I think that redundancy is valuable. This ensures that future models that override the i18n_key will continue to work even if they're not strings, and the `str(self.pk)` change ensures that the default case will use a string no matter what, which helps for consistency in the generated strings before and after they get passed through json

joshlory:
Ref: PR https://github.com/mrjoshida/curriculumbuilder/pull/103.

PR #139, opened 2019-07-11
Keywords proxy

joshlory:
Better approach for PR https://github.com/mrjoshida/curriculumbuilder/pull/118.

Sample Keywords.json entries:

```js
    "collector": {
        "title": "Collector"
    }, 
    "common-sense-education": {
        "title": "Common Sense Education"
    }, 
    "community-session": {
        "title": "community session"
    }, 
    "computer-lab": {
        "title": "computer lab"
    }, 
```

### DIFF
--- a/i18n/management/utils.py
+++ b/i18n/management/utils.py
@@ -8,17 +8,19 @@ from django.utils.translation import to_locale
 
 from django_slack import slack_message
 
-from i18n.models import Internationalizable, InternationalizablePage
+from i18n.models import Internationalizable
 
 
 def should_sync_model(model):
     """
     whether or not the given Django model is something the i18n sync should
     process. We care about models that:
-        - directly extend the Internationalizable or InternationalizablePage
-          model defined by this module
+        - extend the Internationalizable model defined by this module
+        - are not proxy models (unless the model explicitly opts in to translation)
     """
-    return Internationalizable in model.__bases__ or InternationalizablePage in model.__bases__
+    is_internationalizable = issubclass(model, Internationalizable)
+    should_skip = model._meta.proxy and not hasattr(model, 'translate_proxy') # pylint: disable=protected-access
+    return is_internationalizable and not should_skip
 
 def get_models_to_sync():
     """Retrieve all models that should be processed by the i18n sync"""
--- a/i18n/management/utils.py
+++ b/i18n/management/utils.py
@@ -20,7 +20,7 @@ def should_sync_model(model):
     """
     is_internationalizable = issubclass(model, Internationalizable)
     is_not_proxy = not model._meta.proxy # pylint: disable=protected-access
-    return is_internationalizable and is_not_proxy
+    return is_internationalizable
 
 def get_models_to_sync():
     """Retrieve all models that should be processed by the i18n sync"""
--- a/i18n/management/utils.py
+++ b/i18n/management/utils.py
@@ -19,7 +19,7 @@ def should_sync_model(model):
         - are not proxy models (unless the model explicitly opts in to translation)
     """
     is_internationalizable = issubclass(model, Internationalizable)
-    should_skip = model._meta.proxy and not hasattr(model, 'translate_proxy') # pylint: disable=protected-access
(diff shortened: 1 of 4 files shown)


i18n/management/utils.py
```diff
@@ -8,19 +8,17 @@
...
     process. We care about models that:
-        - extend the Internationalizable model defined by this module
-        - are not proxy models (used by Django Admin for editing)
+        - directly extend the Internationalizable or InternationalizablePage
+          model defined by this module
     """
-    is_internationalizable = issubclass(model, Internationalizable)
-    is_not_proxy = not model._meta.proxy # pylint: disable=protected-access
-    return is_internationalizable and is_not_proxy
+    return Internationalizable in model.__bases__ or InternationalizablePage in model.__bases__
```
Hamms: I don't love this approach; what if we instead change the "is not proxy" rule to just be "does not extend the Django Admin model"

lessons/models.py
```diff
@@ -47,6 +50,31 @@
...
                   'strike-through', 'text-align', 'text-decoration']
 
+
+"""
+Keyword
+
+"""
+
+
+class Keyword(BaseKeyword, Internationalizable):
```
Hamms: I'd love to see some extensive comments explaining both what this model and what the `translated_keywords` methods are for, and why they're necessary

i18n/management/utils.py
joshlory: What's the proper way to detect this? Something like `UnitStandard` is an admin proxy model, but only inherits from the `Unit` model:

https://github.com/mrjoshida/curriculumbuilder/blob/57cad8f62e50b03e4f16bf77cd9e2e1da5c3e44e/curricula/admin.py#L86-L90

Hamms: Oh, good point.

Hmm, I'm not sure what's the best way to do this, then. Would something like a `translate_proxy` flag be too gross? Something like

```python
def should_translate(model):
  if not issubclass(model, Internationalizable):
    return False

  if model._meta.proxy:
    return model._meta.translate_proxy or False

  return True
```

joshlory: We could, but I'm not sure I like that better than checking if it it's a direct subclass of one of the I18n models?

Hamms: I really don't like the idea of making _direct_ subclass into a thing that matters; seems then like we want mixins or some implementation that works outside of the class inheritance hierarchy.

To me, it seems like this logic is consistent with Django's underlying philosophy. Proxy models are virtual models that shouldn't matter, so we internationalize everything except proxy models, except we also allow individual proxy models to override that default behavior if there's a special reason.

joshlory: Updated in commit 242912797d7a0377f65e666af4008396e921c9d7! I couldn't add it to the class's Meta because I got:

```
TypeError: 'class Meta' got invalid attribute(s): translate_proxy
```

lessons/models.py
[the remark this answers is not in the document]
joshlory: Renamed to `I18nKeyword` for clarity (in commit d773e99a94f26164a9629f731adcbca6269aae35).

i18n/management/utils.py
```diff
@@ -16,11 +16,11 @@ def should_sync_model(model):
     whether or not the given Django model is something the i18n sync should
     process. We care about models that:
         - extend the Internationalizable model defined by this module
-        - are not proxy models (used by Django Admin for editing)
+        - are not proxy models (unless the model explicitly opts in to translation)
     """
     is_internationalizable = issubclass(model, Internationalizable)
-    is_not_proxy = not model._meta.proxy # pylint: disable=protected-access
-    return is_internationalizable and is_not_proxy
+    should_skip = model._meta.proxy and not hasattr(model, 'translate_proxy') # pylint: disable=protected-access
```
Hamms: should this be 
```suggestion
    should_skip = model._meta.proxy and not getattr(model, 'translate_proxy', False) # pylint: disable=protected-access
```

PR #140, opened 2019-07-18
Translate miscellaneous remaining strings

joshlory:
Hopefully the last few UI strings to add to the `django.po` file!


joshlory:
Added a bunch more strings, but skipping re-review so I can get these into the build tonight.

PR #141, opened 2019-07-19
Translate all standards, not just cross-curricular opportunities

joshlory:
Fixes AP and ETS standards on this view:

image

### DIFF
--- a/standards/models.py
+++ b/standards/models.py
@@ -182,7 +182,7 @@ class Standard(Internationalizable):
 
     @property
     def should_be_translated(self):
-        return any(lesson.should_be_translated for lesson in self.opportunities.all())
+        return True
 
     @classmethod
     def internationalizable_fields(cls):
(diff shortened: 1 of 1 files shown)


joshlory:
Yeah, I'm opting to add everything for now. I haven't found how NGSS are connected to lessons yet.

Hamms:
Do we just need to check all three of these relations? https://github.com/mrjoshida/curriculumbuilder/blob/567ee8b9dbf55c3c5534b04af098968a6e218e4e/lessons/models.py#L326-L330

I'd strongly prefer to maintain the "only sync to CB the things we know we want to translate" rule that we've been following so far.

Hamms:
@joshlory made the changes for https://codedotorg.atlassian.net/browse/FND-613 here rather than opening a new PR.

Captured the resulting content changes here: https://gist.github.com/Hamms/104b2fc5a09885fdbe6673290af4c12e

PR #144, opened 2019-08-08
Only allow certain domains to be used as the src attribute for iframes

Hamms:
https://codedotorg.atlassian.net/browse/FND-291

### DIFF
--- a/curriculumBuilder/settings.py
+++ b/curriculumBuilder/settings.py
@@ -10,6 +10,8 @@ import dj_database_url
 
 from django.utils.translation import ugettext_lazy as _
 
+from urlparse import urlparse
+
 BASE_DIR = os.path.dirname(os.path.dirname(__file__))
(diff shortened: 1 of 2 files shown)


curriculumBuilder/settings.py
```diff
@@ -421,26 +423,55 @@
...
+def iframe_filter_domain(name, value):
+    """
+    We use iframes in curriculum descriptions to embed Google Docs containing
+    the lesson plans. We want to continue to allow iframes in general, but
+    restrict to only allow google docs.
+
+    If in the future we want to expand our usage of iframes to include other use
+    cases, those should be simililarly captured here.
+    """
+    if name in ('width', 'height', 'frameborder', 'style'):
```
bethanyaconnor: For my own education, why is this necessary?

Hamms: this function will override the '*' allowed tags, so we have to make sure that the attributes we want to allow are explicitly allowed here

PR #154, opened 2019-09-16
Test curricula rendering routes

davidsbailey:
Add a few tests which render various views under `curricula/`. This PR does the minimum in each test needed to get code coverage numbers up, which is an important metric for when we are ready to move to python 3.

This increases our coverage within `curricula/` from 4% to 57%, and increases our overall coverage from 17% to 40%.

New coverage report: [cb-coverage-report-02.txt](https://github.com/mrjoshida/curriculumbuilder/files/3618099/cb-coverage-report-02.txt)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -24,4 +24,4 @@ before_script:
   - python manage.py migrate
 
 script:
-  - python manage.py test
+  - debug=true python manage.py test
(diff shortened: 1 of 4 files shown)


.travis.yml
```diff
@@ -24,4 +24,4 @@ before_script:
   - python manage.py migrate
 
 script:
-  - python manage.py test
+  - debug=true python manage.py test
```
madelynkasula: is this change intentional/needed?

davidsbailey: yep - for some reason this is needed for any of the tests which render curricula views to pass on travis.

PR #155, opened 2019-09-16
Clean up between i18n tests

davidsbailey:
A small tweak to i18n tests allows them to pass when the `--keepdb` flag is specified, which reduces the amount of time it takes a developer to run all unit tests locally from 1m23s to 23s.

```
(cb) Dave-MBP:~/src/cb (keepdb-i18n)$ time debug=true ./manage.py test --keepdb
Using existing test database for alias 'default'...
...
Ran 933 tests in 5.102s
OK
Preserving test database for alias 'default'...

real	0m23.131s
user	0m21.181s
sys	0m2.393s
```

Test cases which derive from Django's `TestCase` already wrap DB activity in a transaction, eliminating the need to explicitly clean up.

### DIFF
--- a/i18n/tests/test_internationalizable.py
+++ b/i18n/tests/test_internationalizable.py
@@ -10,8 +10,10 @@ class InternationalizableTestCase(TestCase):
     Basic tests for the Internationalizable abstract model
     """
     def setUp(self):
-        TestModel.objects.all().delete()
         self.test_model = TestModel(id=1, title="Test Title", description="Test Description").save()
+
+    def tearDown(self):
+        TestModel.objects.all().delete()
 
     def test_gather_strings(self):
         """
--- a/i18n/tests/test_internationalizable.py
+++ b/i18n/tests/test_internationalizable.py
@@ -10,7 +10,8 @@ class InternationalizableTestCase(TestCase):
     Basic tests for the Internationalizable abstract model
     """
     def setUp(self):
-        TestModel(title="Test Title", description="Test Description").save()
+        TestModel.objects.all().delete()
+        self.test_model = TestModel(id=1, title="Test Title", description="Test Description").save()
 
     def test_gather_strings(self):
         """


i18n/tests/test_internationalizable.py
```diff
@@ -10,7 +10,8 @@ class InternationalizableTestCase(TestCase):
     Basic tests for the Internationalizable abstract model
     """
     def setUp(self):
-        TestModel(title="Test Title", description="Test Description").save()
+        TestModel.objects.all().delete()
```
bethanyaconnor: Why does this help?

davidsbailey: Good question! Without this, the second test run with `--keepdb` fails, because the data generated in the first run still exists in the database:
```
./manage.py test i18n.tests --keepdb 
...
======================================================================
FAIL: test_gather_strings (i18n.tests.test_internationalizable.InternationalizableTestCase)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/Users/dsb/src/cb/i18n/tests/test_internationalizable.py", line 24, in test_gather_strings
    "title": u"Test Title"
AssertionError: {'1':[56 chars]tle'}, '3': {'description': u'Test Description[89 chars]le'}} != {'1':[56 chars]tle'}}
- {'1': {'description': u'Test Description', 'title': u'Test Title'},
?                                                                   ^

+ {'1': {'description': u'Test Description', 'title': u'Test Title'}}
?                                                                   ^

-  '2': {'description': u'Test Description', 'title': u'Test Title'},
-  '3': {'description': u'Test Description', 'title': u'Test Title'}}
----------------------------------------------------------------------
```

Hamms: I believe this should instead go [in the `tearDown` method](https://docs.python.org/3/library/unittest.html#unittest.TestCase.tearDown)

PR #157, opened 2019-09-24
Invoke CodeMirror later on certain pages

islemaster:
[LP-287](https://codedotorg.atlassian.net/browse/LP-287): We're seeing an issue where CodeMirror is applying incorrect spacing to HTML snippets only when the page is rendered within an iframe on a Code Studio map level.  This seems to be due to a timing issue - the CodeMirror handler is being invoked before the document in the iframe has fully loaded, leading to some incorrect calculations.

| [Embedded](https://studio.code.org/s/csd2-2018/stage/6/puzzle/6) | [Not Embedded](https://studio.code.org/docs/concepts/html/lists/index.html) |
| --- | --- |
| image | image |

Here I'm suggesting we use jQuery's `$(window).on("load", fn)` instead of `$(fn)` / `$(document).ready(fn)` which explicitly waits for the whole page (including iframes) to be ready, not just the DOM.

Read more: https://learn.jquery.com/using-jquery-core/document-ready/

I'm not sure of the quickest way to verify this change.  Recommendations please?

### DIFF
--- a/documentation/templates/documentation/block.html
+++ b/documentation/templates/documentation/block.html
@@ -94,7 +94,7 @@
 <script src="{% static "codemirror/mode/javascript/javascript.js" %}"></script>
 <script src="{% static "codemirror/mode/htmlmixed/htmlmixed.js" %}"></script>
 <script type="text/javascript">
-$(function() {
+$(window).on("load", function() {
   $('pre').each(function() {
 
         var $this = $(this);
--- a/documentation/templates/documentation/embed.html
+++ b/documentation/templates/documentation/embed.html
@@ -39,7 +39,7 @@
 <script src="{% static "codemirror/mode/javascript/javascript.js" %}"></script>
 <script src="{% static "codemirror/mode/htmlmixed/htmlmixed.js" %}"></script>
 <script type="text/javascript">
-$(function() {
+$(window).on("load", function() {
   $('pre').each(function() {
 
         var $this = $(this);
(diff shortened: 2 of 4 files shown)


islemaster:
Oh, the eyes test is a good idea.  I'll add one.

davidsbailey:
regarding process, did you have to open up this PR like a contributor would @islemaster ?
Screen Shot 2019-09-27 at 10 20 13 AM

seems ok if this is a transitional thing, but if there anything preventing you from opening up future pull requests the way we do in the code-dot-org repo let's get those addressed.

islemaster:
Yeah, I had to use fork-and-pull while the repo was owned by @mrjoshida.  As of its move into the code-dot-org organization today I have permission to push branches directly.

davidsbailey:
> I'm not sure of the quickest way to verify this change. Recommendations please?

one idea for manual verification would be to add a large image to one of the affected pages, and turn on network throttling. If that is successful in reproducing the problem while using $().ready, then you can test whether switching to `onload` is an effective solution.

davidsbailey:
Just a heads up that merging this (unless another PR goes in first) is likely to cause the first deploy since moving things over from github.com/mrjoshida . so it may make sense to do so at a time when we can respond to any issues

islemaster:
Oh!  Thanks for the heads-up, I didn't think of that.  I'll save it for next week.

PR #164, opened 2019-10-22
Log all errors from manage.py commands

Hamms:
Primarily so we can make sure to capture errors in the i18n sync, which is implemented as an automated series of manage.py commands. But also generally is just good practice.

### DIFF
--- a/curriculumBuilder/settings.py
+++ b/curriculumBuilder/settings.py
@@ -741,6 +741,11 @@ LOGGING = {
             'handlers': ['console', 'mail_admins', 'slack_admins'],
             'level': 'DEBUG',
             'propagate': True
+        },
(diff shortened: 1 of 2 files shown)


Hamms:
Yep!

The line `'handlers': ['console', 'slack_admins'],` tells the logger to use these two handlers:

https://github.com/code-dot-org/curriculumbuilder/blob/298d23bebe952edf7091dc745ce35783e2e70dde/curriculumBuilder/settings.py#L683-L686

and 

https://github.com/code-dot-org/curriculumbuilder/blob/298d23bebe952edf7091dc745ce35783e2e70dde/curriculumBuilder/settings.py#L692-L697

which will of course log the message to the console and to our slack channel, respectively.

PR #165, opened 2019-10-22
don't attempt to gather strings if there are no objects to gather from

Hamms:
Partially to save time, partially so we don't attempt to divide by zero later: https://github.com/code-dot-org/curriculumbuilder/blob/4fa9f93c7c6637c9fae3dcdceaa01a79c7f33873/i18n/models.py#L75


Hamms:
Thanks!

PR #167, opened 2019-10-22
improve i18n sync error messaging

Hamms:
Previously, we just logged each step of the i18n sync individually, including an indication that all steps had finished. The problem with that is of course that if all steps _didn't_ finish, you have to be paying explicit attention to notice.

This adds a more explicit error message when anything goes wrong.

### DIFF
--- a/i18n/management/commands/log_i18n_failure.py
+++ b/i18n/management/commands/log_i18n_failure.py
@@ -9,4 +9,4 @@ class Command(BaseCommand):
     See bin/sync_i18n.sh for context.
     """
     def handle(self, *args, **options):
-        log("WARNING: I18n Sync encountered an error")
+        log("ERROR: I18n Sync encountered a problem and did not complete")
--- a/i18n/management/commands/log_i18n_failure.py
+++ b/i18n/management/commands/log_i18n_failure.py
@@ -1,0 +1,12 @@
+# pylint: disable=missing-docstring
+from i18n.management.utils import log
+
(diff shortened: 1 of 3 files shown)


i18n/management/commands/log_i18n_failure.py
```diff
@@ -0,0 +1,12 @@
...
+
+from django.core.management.base import BaseCommand
+
+class Command(BaseCommand):
+    """
+    Used by the i18n sync to alert that not all steps finished successfully.
+    See bin/sync_i18n.sh for context.
+    """
+    def handle(self, *args, **options):
+        log("WARNING: I18n Sync encountered an error")
```
daynew: Minor comments mostly out of my own python ignorance:
* Why is the "error" logged as a "WARNING" instead of an "ERROR"? Could everything be okay if there was an error?
* Is there a 'error(...)' or 'warn(...)' function which might be more appropriate?

Hamms: 1. No reason at all! I like "ERROR" better, I'll update it.
2. Not right now; the `log` method we're using here is defined at https://github.com/code-dot-org/curriculumbuilder/blob/54f9f52f686eea0909d73c3b7aadb6553f23b949/i18n/management/utils.py#L49-L58 and is pretty minimal. What we really _should_ be using is an actual python logger like we are in https://github.com/code-dot-org/curriculumbuilder/pull/164 but I'm sticking with what we already have for this change

PR #168, opened 2019-10-23
Filter vocab admin view for partners to those owned by current user 

davidsbailey:
## Description

partially addresses [LP-829](https://codedotorg.atlassian.net/browse/LP-829). Limits access for anyone without the `access_all_vocab` permission to only be able to view, change or delete vocabs which they own. the migrations which set us up for this will be ship ahead of this PR in #169. 

This solution relies on the [Ownable](http://mezzanine.jupo.org/docs/packages.html#mezzanine.core.models.Ownable) and [OwnableAdmin](http://mezzanine.jupo.org/docs/packages.html#mezzanine.core.admin.OwnableAdmin) classes in mezzanine, which is the CMS currently used by django to manage page hierarchy. It also relies on the existing user/group permissions model built into django.

This breaks down as follows:
* `Ownable` assigns every object an owner, which by default is whoever created it
* `OwnableAdmin` filters the admin view to only show the objects owned by the current user
* `FilterableAdmin` extends `OwnableAdmin` to skip that filtering for users with the "access_all" property

Side note: this solution assumes that all curriculum writers from Broward will share one account on Curriculum Builder. 

## screenshots

A new permission now shows up in the admin page for editing a user. I've assigned this to my user account `Dave` locally:
Screen Shot 2019-10-23 at 8 25 25 AM

As a result, I can see all vocabs:
Screen Shot 2019-10-23 at 8 26 31 AM

Our partner, `Broward`, on the other hand, does not have the access_all_vocab permission:
Screen Shot 2019-10-23 at 8 27 39 AM

When Broward goes to the admin page for vocab, they can only see vocab words they created:
Screen Shot 2019-10-23 at 8 32 09 AM

## Deployment plan
This sequence ensures that Broward's permissions will be restricted, without interrupting any authors' ability to edit vocab words:
- [ ] ship migrations in #169
- [ ] add the `access_all_vocab` property to the `author` group, so that authors do not lose ability to access all vocab after the following step
- [ ] ship this PR

## Future work
next steps are to ship similar changes for other fields which Broward needs to edit:
Screen Shot 2019-10-21 at 2 40 52 PM

### DIFF
--- a/lessons/admin.py
+++ b/lessons/admin.py
@@ -6,7 +6,7 @@ from django.forms import TextInput, ModelForm
 from import_export import resources, fields
 from import_export.admin import ImportExportModelAdmin
 from import_export.widgets import ManyToManyWidget, ForeignKeyWidget
-from mezzanine.core.admin import StackedDynamicInlineAdmin, TabularDynamicInlineAdmin
+from mezzanine.core.admin import StackedDynamicInlineAdmin, TabularDynamicInlineAdmin, OwnableAdmin
 from mezzanine.generic.fields import KeywordsField
 from mezzanine.pages.admin import PageAdmin
 from reversion.admin import VersionAdmin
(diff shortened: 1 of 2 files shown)


davidsbailey:
any input, @Hamms ?

PR #169, opened 2019-10-23
Vocab migrations

davidsbailey:
migrations for https://github.com/code-dot-org/curriculumbuilder/pull/168 . See that PR for context.

### DIFF
--- a/lessons/migrations/0044_auto_20191021_1519.py
+++ b/lessons/migrations/0044_auto_20191021_1519.py
@@ -1,0 +1,32 @@
+# -*- coding: utf-8 -*-
+from __future__ import unicode_literals
+
+from django.db import models, migrations
+from django.conf import settings
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('generic', '0002_auto_20141227_0224'),
+        migrations.swappable_dependency(settings.AUTH_USER_MODEL),
+        ('lessons', '0043_auto_20190520_1630'),
+    ]
+
+    operations = [
+        migrations.CreateModel(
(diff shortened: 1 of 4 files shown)


standards/migrations/0007_auto_20191021_1519.py
```diff
@@ -0,0 +1,19 @@
...
+
+    dependencies = [
+        ('standards', '0006_auto_20170929_1528'),
+    ]
+
+    operations = [
+        migrations.AlterField(
+            model_name='standard',
+            name='gradeband',
+            field=models.ForeignKey(related_name='standards', to='standards.GradeBand'),
```
davidsbailey: this change appears to be a side effect of running `./manage.py makemigrations` without any of the changes listed in this PR. I'm including it here since it seems like otherwise it'll eventually get included in whatever other PR tries to make migrations next.

lessons/migrations/0044_auto_20191021_1519.py
```diff
@@ -0,0 +1,32 @@
...
+            ],
+            options={
+                'proxy': True,
+            },
+            bases=('generic.keyword', models.Model),
+        ),
+        migrations.AddField(
+            model_name='vocab',
+            name='user',
+            field=models.ForeignKey(related_name='vocabs', default=1, verbose_name='Author', to=settings.AUTH_USER_MODEL),
```
davidsbailey: `default=1` assigns all existing vocabs to be owned by the root user:
```
In [1]: Vocab.objects.all().first().user
Out[1]: <User: root>
```

PR #172, opened 2019-10-25
Prevent partners from editing objects they do not own

davidsbailey:
The goal is to allow partners such as Broward to have the following permissions:
Screen Shot 2019-10-21 at 2 40 52 PM

 and to limit those permissions so that they can only edit objects they own. See https://github.com/code-dot-org/curriculumbuilder/pull/168 including screenshots for how we accomplished this for Vocab objects using Mezzanine's `Ownable` and `OwnableAdmin` classes, and the new `access_all_` permission on each model type. The same solution, added in the first 2 commits of this PR, works for other objects which show up in the admin view as a list, namely `Resource`, `Activity`, and `Objective`:
broward-activity-objective-resource

Objects which inherit from Mezzanine's [Page model](http://mezzanine.jupo.org/docs/content-architecture.html#the-page-model) work differently, though. All `Page` types (`Curriculum`, `Chapter`, `Unit` and `Lesson`) are shown together in the admin Page view, and the `OwnableAdmin` class fails to filter them out of what appears there. Instead, we define `can_add`, `can_move`, `can_change` and `can_delete` (documented [here](http://mezzanine.jupo.org/docs/content-architecture.html#page-permissions)) to restrict which controls are available in that view. The result is the partner can freely rearrange their own Pages:
broward-page-yes

but they receive an error and a page reload if they try to mix their pages with ones they don't own:
broward-page-no

## Deployment plan
This sequence ensures that Broward's permissions will be restricted, without preventing authors or moderators from edit these objects at any point during the deployment:
- [x] ship migrations in https://github.com/code-dot-org/curriculumbuilder/pull/173
- [x] add the `access_all_` property for each model type to the `author` and `moderator` groups
- [ ] ship this PR

## Future work
1. limit what shows up in the "Lesson" dropdown when editing Objectives or Activities (see top anigif), to only include Lessons owned by the partner
2. figure out how to test this

### DIFF
--- a/curricula/admin.py
+++ b/curricula/admin.py
@@ -55,7 +55,7 @@ class CurriculumAdmin(PageAdmin, VersionAdmin, FilterableAdmin):
     inlines = (TopicInline,)
 
     def can_access_all(self, request):
-        return request.user.has_perm('curriculum.access_all_curricula')
+        return request.user.has_perm('curricula.access_all_curricula')
 
 
 
--- a/curricula/admin.py
+++ b/curricula/admin.py
@@ -8,7 +8,7 @@ from import_export.admin import ImportExportModelAdmin
 from reversion.admin import VersionAdmin
 
 from lessons.models import Lesson
-from lessons.admin import LessonForm
+from lessons.admin import LessonForm, FilterableAdmin
 
 from curricula.models import Curriculum, Unit, Chapter, Topic
 from standards.models import Standard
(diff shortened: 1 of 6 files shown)


curricula/admin.py
```diff
@@ -48,19 +48,26 @@ class TopicInline(TabularDynamicInlineAdmin):
...
 
-class CurriculumAdmin(PageAdmin, VersionAdmin):
+class CurriculumAdmin(PageAdmin, VersionAdmin, FilterableAdmin):
     model = Curriculum
     verbose_name_plural = "Curricula"
     filter_horizontal = ('frameworks',)
     inlines = (TopicInline,)
 
+    def can_access_all(self, request):
+        return request.user.has_perm('curriculum.access_all_curricula')
```
Hamms: should this be

```suggestion
        return request.user.has_perm('curricula.access_all_curricula')
```

?

curricula/models.py
```diff
@@ -71,10 +74,24 @@ def internationalizable_fields(cls):
...
+        return self.can_access(request)
+
+    def can_delete(self, request):
+        return self.can_access(request)
+
     def can_move(self, request, new_parent):
         if new_parent is not None:
             msg = 'Curriculum must be a top level object'
             raise PageMoveException(msg)
+        if not request.user.has_perm('curricula.access_all_curricula'):
```
Hamms: is there a reason this isn't using `can_access`?

davidsbailey: Yes! We do not want the partner to be able to move their `Curriculum` around within admin Page view shown in the anigifs, even if they own it. I've added a comment.

davidsbailey:
Unit tests forthcoming! https://codedotorg.atlassian.net/browse/LP-919

PR #174, opened 2019-10-25
add linux-friendly database config to local settings example

davidsbailey:
these changes are needed in local_settings.py in order for Linux setup instructions to work, and adding them doesn't seem to break running the local server on Mac, so rather than adding another step to the setup it seems best to just include these in the local settings file which is already copied into place during existing setup steps.

### DIFF
--- a/curriculumBuilder/local_settings.py.example
+++ b/curriculumBuilder/local_settings.py.example
@@ -21,6 +21,6 @@ CACHES = {
 DATABASES = {
     "default": {
         'ENGINE': 'django.db.backends.postgresql_psycopg2',
-	'NAME': 'curriculumbuilder',
+        'NAME': 'curriculumbuilder',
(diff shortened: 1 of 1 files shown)


curriculumBuilder/local_settings.py.example
```diff
@@ -17,3 +17,10 @@ CACHES = {
         'LOCATION': 'unique-snowflake',
     }
 }
+
+DATABASES = {
+    "default": {
+        'ENGINE': 'django.db.backends.postgresql_psycopg2',
+	'NAME': 'curriculumbuilder',
```
Hamms: some weird spacing here, but otherwise LGTM!

davidsbailey: thank you for catching! it was a tab 😱

PR #175, opened 2019-10-25
Limit what partners can select in lesson dropdowns

davidsbailey:
The admin pages for `Objective` and `Activity` have a dropdown to assign the object to a lesson. We want to make sure that partners can only assign these objects to their own lessons. I did not see an easy way to filter what shows up in the menu, but I did figure out how to reject the request if they select a lesson which does not belong to them:
broward-lesson-dropdown

I got the idea to override `save_form` because I saw that that's what the `OwnableAdmin` class overrode to accomplish custom behavior when an object was being updated: http://mezzanine.jupo.org/docs/_modules/mezzanine/core/admin.html#OwnableAdmin

It's not the best UX, because it lets partners select a bad value and then gives an ugly exception. But it will at least keep their data in a good state.

### DIFF
--- a/lessons/admin.py
+++ b/lessons/admin.py
@@ -72,6 +72,13 @@ class ObjectiveAdmin(FilterableAdmin):
     def can_access_all(self, request):
         return request.user.has_perm('lessons.access_all_objectives')
 
+    def save_form(self, request, form, change):
+        lesson_id = form.cleaned_data['lesson'].id
+        lesson = Lesson.objects.get(id=lesson_id)
+        if not lesson.can_access(request):
+            raise Exception('You can only add an objective to a lesson you own')
+        return super(ObjectiveAdmin, self).save_form(request, form, change)
+
 class PrereqInline(StackedDynamicInlineAdmin):
     model = Prereq
     verbose_name = "Prerequisite"
@@ -469,6 +476,13 @@ class ActivityAdmin(CompareVersionAdmin, FilterableAdmin):
 
     def can_access_all(self, request):
         return request.user.has_perm('lessons.access_all_activities')
+
+    def save_form(self, request, form, change):
+        lesson_id = form.cleaned_data['lesson'].id
+        lesson = Lesson.objects.get(id=lesson_id)
+        if not lesson.can_access(request):
+            raise Exception('You can only add an activity to a lesson you own')
+        return super(ActivityAdmin, self).save_form(request, form, change)
 
 
 class AnnotationAdmin(CompareVersionAdmin):


lessons/admin.py
```diff
@@ -470,6 +477,13 @@ def unit(self, instance):
     def can_access_all(self, request):
         return request.user.has_perm('lessons.access_all_activities')
 
+    def save_form(self, request, form, change):
+        lesson_id = form.cleaned_data['lesson'].id
+        lesson = Lesson.objects.get(id=lesson_id)
+        if not lesson.can_access(request):
+            raise Exception('You can only add an activity to a lesson you own')
+        return super(ActivityAdmin, self).save_form(request, form, change)
+
```
Erin007: Why do we need this twice - once for ActivityAdmin and once for ObjectiveAdmin? Maybe extract or consolidate?

Erin007:
Does this only apply to the restricted permissions we're giving to Broward? What if let's say Hannah wants to add an objective to a lesson that GT made, will she be able to still do that?

davidsbailey:
> Does this only apply to the restricted permissions we're giving to Broward? What if let's say Hannah wants to add an objective to a lesson that GT made, will she be able to still do that?

Good question Erin -- yes, our curriculum team will still be able to select any Lesson in these menus. The `can_access` method on the Lesson model grants access to the following users:
* anyone with "access all lessons" permissions (including all authors, which includes Hannah, GT, etc)
* the specific user who created the lesson (e.g. Broward, but only if they have selected a Broward lesson)
https://github.com/code-dot-org/curriculumbuilder/blob/3df69280625fe7313b2a832c5955c16edabf930e/lessons/models.py#L373-L374

I will add some better documentation for this method in the next PR.

lessons/admin.py
davidsbailey: Let's see. I could extract this into a parent class:
```
def validate_form_child(self, request, form, field_name, model):
    id = form.cleaned_data[field_name].id
    object = model.objects.get(id=id)
    if not object.can_access(request)
        raise Exception("%s field must refer to a %s you own" % (field_name, field_name))
```
and then call it like this here:
```
def save_form(self, request, form, change):
    self.validate_form_child(request, form, field_name, model)
    return super(ActivityAdmin, self).save_form(request, form, change)
```
Does that look better? 

I am going to create a parent `Filterable` class in the next PR, so I could extract this in that PR.

Erin007: Ah. Thanks for the explanation. I don't feel strongly, mostly curious.  If it makes sense for follow-up, works for me.

davidsbailey: I was hesitant at first, but yes, I think this is a good follow-up!

PR #180, opened 2019-10-28
extract and implement Filterable class in Ownable models

davidsbailey:
### Background

https://github.com/code-dot-org/curriculumbuilder/pull/173 added `Ownable` as a parent to several models, which broke inline editing because it overrode `is_editable` to ignore our existing `change_*` permissions.  A quick fix for this was made to `Lesson` and `Activity` (removed in this PR), to allow curriculum team members to regain inline editing privileges. However the following two problems remain:
1. partners (who do not have `access_all_*` permissions) **can** currently inline-edit any `Lesson` and `Activity` objects
2. curriculum team (who have `access_all_*` permissions) **cannot** currently inline-edit any `Curriculum` or `Unit` objects

### Description

This PR does the following:
1. extracts a `Filterable` interface which overrides `is_filterable` for all `Ownable` models
2. implements correct logic in `is_filterable` on all `Ownable` models to fix permissions for all user types

I'm stopping here (see Future Work) because these permission fixes are needed for the bug bash at 12:30pm PT on Tuesday.

### Future work

I could not figure out how to share make a class in `curricula.models` inherit from `lessons.models.Filterable`, perhaps due to a circular reference (both model files import the other). Because `Filterable` and `FilterableAdmin` really belong in a shared location, my plan is to extract these to somewhere like `core.models` in a subsequent PR. Ideas for this are welcome!

I also still owe tests for everything Ownable/Filterable.

### DIFF
--- a/lessons/models.py
+++ b/lessons/models.py
@@ -108,7 +108,7 @@ Vocabulary
 """
 
 
-class Vocab(Internationalizable, Ownable):
+class Vocab(Internationalizable, Filterable):
     word = models.CharField(max_length=255)
     simpleDef = models.TextField()
     detailDef = models.TextField(blank=True, null=True)
@@ -147,6 +147,12 @@ class Vocab(Internationalizable, Ownable):
             self.detailDef = self.simpleDef
         super(Vocab, self).save(*args, **kwargs)
 
+    def can_access(self, request):
+        return request.user.has_perm('lessons.access_all_vocab') or request.user.id == self.user_id
+
+    def is_editable(self, request):
+        return self.can_access(request) and request.user.has_perm('lessons.change_vocab')
+
 
 """
 Linked Resources
@@ -154,7 +160,7 @@ Linked Resources
 """
 
 
-class Resource(Orderable, Internationalizable, Ownable):
+class Resource(Orderable, Internationalizable, Filterable):
     name = models.CharField(max_length=255)
     type = models.CharField(max_length=255, blank=True, null=True)
     student = models.BooleanField('Student Facing', default=False)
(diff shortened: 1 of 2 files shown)


davidsbailey:
> I'm wondering about the plan for the issue with editing curriculum and units. Have you communicated that to the curriculum team? Is there a work around?

Hey Dani, this PR fixes the problem with curriculum and units. these will be edited by anyone with the `access_all_*` and `change_*` permission for those object types, which includes all curriculum writers.

davidsbailey:
@Hamms and @dmcavoy I'm going to go ahead and merge (bug bash is actually at 11:30) but please feel free to follow up with any additional feedback and I'll address it in a next PR.

lessons/models.py
```diff
@@ -81,13 +81,34 @@ def translate_proxy(self):
...
+
+    # Helper method which should return true if the object is either owned by
+    # the user, or the user has permission to access all objects of this type.
+    def can_access(self, request):
+        raise NotImplementedError
+
+    # Restrict in-line editing of the object, overriding behavior defined here:
+    # http://mezzanine.jupo.org/docs/_modules/mezzanine/core/models.html#Ownable
+    def is_editable(self, request):
+        raise NotImplementedError
```
Hamms: should these return a reasonable default instead of erroring out? In particular, I can see value in doing something like 

```python
class Filterable
    # Helper method which should return true if the object is either owned by
    # the user, or the user has permission to access all objects of this type.
    def can_access(self, request):
        return request.user.id == self.user_id

    # Restrict in-line editing of the object, overriding behavior defined here:
    # http://mezzanine.jupo.org/docs/_modules/mezzanine/core/models.html#Ownable
    def is_editable(self, request):
        return self.can_access(request)
```

And then in subclasses doing

```python
class Vocab(Internationalizable, Filterable):
    def can_access(self, request):
        return request.user.has_perm('lessons.access_all_vocab') or super(Vocab, self).can_access(request)

    def is_editable(self, request):
        return request.user.has_perm('lessons.change_vocab') and super(Vocab, self).is_editable(request)
```

davidsbailey: I do not think we would want to `return self.can_access(request)` by default because this could unintentionally grant editing permission to someone without `change_*` permissions for a new object type. That might not be terrible, but it would violate the permission rules we've set up.

The idea was for this to be an abstract method which the subclass is required to define. If the subclass fails to implement it, that's a programming error so we raise. I'm not sure if this is the issue but I could document that in the comments if that helps.

PR #182, opened 2019-10-29
Actually limit what partners can select in lesson dropdowns

davidsbailey:
Follow-on to https://github.com/code-dot-org/curriculumbuilder/pull/175 , which I merged into the wrong branch. this just merges that PR into master.


davidsbailey:
FYI @Erin007 . This is why you still saw the problem in the lesson dropdown when editing an activity or objective during our bug bash this morning.

PR #183, opened 2019-10-29
Hide admin controls from partners on pages they do not own

davidsbailey:
In response to bug bash feedback, this PR hides the admin controls in the settings menu for partners who are viewing a page which they do not own. This includes the following buttons: Publish, Publish PDFs, Clone,  Get Code Studio Stage Details.

The Resources and Vocab sections are preserved, because those are useful when viewing lessons you don't own.

## before

lesson | hoc lesson
--- | --- 
Screen Shot 2019-10-29 at 2 38 10 PM | Screen Shot 2019-10-29 at 2 00 53 PM

## after

lesson | hoc lesson
--- | --- 
Screen Shot 2019-10-29 at 2 38 22 PM | Screen Shot 2019-10-29 at 2 01 09 PM

### DIFF
--- a/curricula/views.py
+++ b/curricula/views.py
@@ -248,6 +248,9 @@ def lesson_view(request, slug, unit_slug, lesson_num, optional_num=False):
 
     changelog = Version.objects.get_for_object(lesson).filter(revision__user__username=settings.CHANGELOG_USER)
 
+    # disable admin controls if user cannot edit inline
+    can_administer = lesson.is_editable(request)
+
     if lesson.unit.lesson_template_override:
         template = lesson.unit.lesson_template_override
     else:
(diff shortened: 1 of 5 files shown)


curricula/views.py
```diff
@@ -248,14 +248,18 @@ def lesson_view(request, slug, unit_slug, lesson_num, optional_num=False):
 
     changelog = Version.objects.get_for_object(lesson).filter(revision__user__username=settings.CHANGELOG_USER)
 
+    # disable admin controls if user cannot edit inline
+    can_administer = lesson.is_editable(request)
```
davidsbailey: pass this in because templates can't pass arguments to methods: https://docs.djangoproject.com/en/1.8/ref/templates/language/#accessing-method-calls

davidsbailey:
Sounds good to me! Updated screenshots:

lesson | hoc lesson
--- | --- 
Screen Shot 2019-10-30 at 11 56 55 AM | Screen Shot 2019-10-30 at 11 56 42 AM

PR #184, opened 2019-10-29
extract factory classes from curricula rendering tests

davidsbailey:
extracting factories for django model classes via [factory boy](https://factoryboy.readthedocs.io/en/latest/introduction.html) turned out to be a piece of cake once I realized that I needed to use their `factory.django.DjangoModelFactory` class instead of their more general `factory.Factory` class. I mostly followed these recipes: https://factoryboy.readthedocs.io/en/latest/recipes.html one gotcha is also documented inline.

This should set us up well to add more tests around permissions without so much setup code bloat.

### DIFF
--- a/curricula/factories.py
+++ b/curricula/factories.py
@@ -1,7 +1,17 @@
-from factory import Sequence
+from django.contrib.auth.models import User
+
+from factory import Sequence, PostGenerationMethodCall
 from factory.django import DjangoModelFactory
 
 from curricula.models import Curriculum
+
+
+class UserFactory(DjangoModelFactory):
+    class Meta:
+        model = User
+
+    username = Sequence(lambda n: "user_%d" % n)
+    password = PostGenerationMethodCall('set_password', 'password')
 
 class CurriculumFactory(DjangoModelFactory):
     class Meta:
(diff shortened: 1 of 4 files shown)


curricula/factories.py
```diff
@@ -0,0 +1,32 @@
...
+
+from curricula.models import Curriculum, Unit
+
+
+class UserFactory(DjangoModelFactory):
+    class Meta:
+        model = User
+
+    username = Sequence(lambda n: "user_%d" % n)
+    password = PostGenerationMethodCall('set_password', 'password')
```
davidsbailey: borrowed from https://stackoverflow.com/questions/15616277/how-can-you-create-an-admin-user-with-factory-boy/17980414#17980414

curricula/tests/test_curricula_rendering.py
```diff
@@ -1,95 +1,46 @@
...
 class CurriculaRenderingTestCase(TestCase):
     def setUp(self):
-        user = User.objects.create_user(username='admin', password='12345')
-        self.user = user
+        user = UserFactory()
         user.is_staff = True
         user.save()
-        self.client.login(username='admin', password='12345')
+        self.user = user
+        self.client.login(username=user.username, password='password')
```
davidsbailey: use the raw password, because `user.password` is encrypted

PR #185, opened 2019-10-30
send staff users to /admin on 403

davidsbailey:
This makes it so that when Broward clicks these links on pages they don't own, they'll be taken to codecurricula.com/admin instead of studio.code.org/courses:
Screen Shot 2019-10-30 at 11 25 12 AM

docs on django error views can be found here: https://docs.djangoproject.com/en/2.2/ref/views/#error-views but basically the existing 403.html file is the default location for the error template that gets rendered when a permission denied error occurs. `request` is the only accessible object in that view, unless you define a custom `handler403` view.

### DIFF
--- a/templates/403.html
+++ b/templates/403.html
@@ -7,6 +7,7 @@
         {% if request.user.is_staff %}
             window.location = "/admin";
         {% else %}
+            {% comment %} avoid redirect loops for non-staff users {% endcomment %}
             window.location = "https://code.org/educate";
         {% endif %}
     </script>
--- a/templates/403.html
+++ b/templates/403.html
@@ -7,7 +7,7 @@
         {% if request.user.is_staff %}
             window.location = "/admin";
         {% else %}
-            window.location = "https://code.org/learn";
+            window.location = "https://code.org/educate";
         {% endif %}
     </script>
 </head>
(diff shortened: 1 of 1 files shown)


templates/403.html
```diff
@@ -4,7 +4,11 @@
     <meta charset="UTF-8">
     <title>403</title>
     <script type="text/javascript">
-    window.location = "https://code.org/educate";
+        {% if request.user.is_staff %}
+            window.location = "/admin";
+        {% else %}
+            window.location = "https://code.org/learn";
```
davidsbailey: I didn't find a way to test with a non-staff user, since non-staff users already get redirected to login on any admin page. so this else clause is a somewhat speculative measure to keep non-staff users (perhaps if we ever fix the "site" option on the login page) from getting stuck in a redirect loop.

templates/403.html
```diff
@@ -4,7 +4,11 @@
     <meta charset="UTF-8">
     <title>403</title>
     <script type="text/javascript">
-    window.location = "https://code.org/educate";
+        {% if request.user.is_staff %}
+            window.location = "/admin";
+        {% else %}
+            window.location = "https://code.org/educate";
```
davidsbailey: I didn't find a way to test with a non-staff user, since non-staff users already get redirected to login on any admin page. so this else clause is a somewhat speculative measure to keep non-staff users (perhaps if we ever fix the "site" option on the login page) from getting stuck in a redirect loop.

PR #187, opened 2019-10-30
Filter privileged resource fields

davidsbailey:
# Background

Yesterday's bug bash revealed that there are many fields we want to hide from platformization partners such as Broward. 

# Description

Hide the Force I18n field from the Resource admin page for users without the `access_all_resources` permission. As noted in comments, this permission now serves as a proxy for what might better be named "administer resources". The plan is to hold off on renaming it until we see what our other needs might be as Broward starts to work with Curriculum Builder.

## Links

The following docs show APIs related to dynamic fields and fieldsets:
https://docs.djangoproject.com/en/1.8/ref/contrib/admin/#django.contrib.admin.ModelAdmin.get_fields
https://docs.djangoproject.com/en/1.8/ref/contrib/admin/#django.contrib.admin.ModelAdmin.get_fieldsets

Unfortunately, certain related methods such as [get_exclude](https://docs.djangoproject.com/en/1.11/ref/contrib/admin/#django.contrib.admin.ModelAdmin.get_exclude) won't be available until we upgrade to at least Django 1.11, which happens to be the last version of Django which supports Python 2.

## Testing story

New unit tests verify that the privileged field is hidden for those without `access_all_resources` permissions.

# Reviewer Checklist:

- [x] Tests provide adequate coverage
- [x] Code is well-commented
- [x] New features are translatable or updates will not break translations
- [x] Relevant documentation has been added or updated
- [x] User impact is well-understood and desirable

### DIFF
--- a/lessons/tests.py
+++ b/lessons/tests.py
@@ -1,3 +1,0 @@
-from django.test import TestCase
-
-# Create your tests here.
--- a/lessons/admin.py
+++ b/lessons/admin.py
@@ -436,10 +436,12 @@ class ResourceAdmin(AjaxSelectAdmin, ImportExportModelAdmin, FilterableAdmin):
     list_editable = ('type', 'student', 'gd', 'url', 'dl_url')
     list_filter = ('lessons__curriculum',)
     inlines = [LessonResourceInline]
-    fields = ['name', 'type', 'student', 'gd', 'url', 'dl_url', 'slug', 'force_i18n']
 
     def can_access_all(self, request):
         return request.user.has_perm('lessons.access_all_resources')
+
+    def get_fields(self, request, obj=None):
+        return ['name', 'type', 'student', 'gd', 'url', 'dl_url', 'slug', 'force_i18n']
 
 
 class VocabAdmin(ImportExportModelAdmin, FilterableAdmin):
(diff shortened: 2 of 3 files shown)


davidsbailey:
## screenshots

Screen Shot 2019-10-30 at 2 59 15 PM

davidsbailey:
sorry, wrong screenshot the first time:

## Screenshots

filter-resource-admin

lessons/tests.py
```diff
@@ -1,3 +0,0 @@
-from django.test import TestCase
-
-# Create your tests here.
```
dmcavoy: Did you delete this on purpose?

davidsbailey:
> Any idea why this file (lessons/tests/init.py) is showing up in the files changed list but says it has zero changes?

I added this file with no content. without doing this, the test command `./manage.py test` won't find the test files in that directory. 

According to https://pythontips.com/2013/07/28/what-is-__init__-py/ , this file is needed by python in order to be able to import things that are in that directory. 

I haven't been able to find any django-specific documentation on why this file is needed in the `tests/` directory, but I would assume based on the python docs that django ends up looking for and importing those test files and that it cannot import them without `__init__.py`.

lessons/tests.py
davidsbailey: Yep -- the guidance as i understand it is to split tests out into multiple files in a `tests/` directory once you have enough of them. we only have one test file here right now, but the goal is to have more. some details here: https://docs.djangoproject.com/en/2.2/topics/testing/overview/#writing-tests

PR #211, opened 2019-12-12
Add testserver as a host

dmcavoy:
When I tried to publish a map I got a bunch of no know host errors that said the host was testserver (https://codedotorg.slack.com/archives/C0SUN2VT5/p1576181556149800). Going to add testserver here and see if that fixes things.


davidsbailey:
Yay, logging worked! LGTM

PR #210, opened 2019-12-11
Log host

dmcavoy:
# Description

Issue: 
- Go to : https://studio.code.org/s/csd4-2019/stage/12/puzzle/13
- Go to help and tips tab
- Click on Design Mode
- In pop up right click on one of the things in the left nav bar
- Page gives error
- In url add `docs/` in front on concepts
- No more page error

There have been previous PRs trying to fix this change:
Previous attempt to fix which did not work: https://github.com/code-dot-org/curriculumbuilder/pull/203  (This PR reverts that change). 

In addition #204 , #205 and #206 implemented most of these changes for the first time but I reverted them out because I was getting `/documentation/concepts/None` for the URL for the menu items in the concepts menu. We think it might be because the host is www.codecurricula.com instead of codecurricula.com so I changed that here. I also added a case to `get_absolute_url_for_host` if the host does not match any of the provided hosts we will default back to the slug and log an error to the console and slack about what the name of the host is.

## Links

- [jira](https://codedotorg.atlassian.net/browse/LP-910)

## Testing story

Unit tests for `get_absolute_url_for_host` and `get_children`.

# Reviewer Checklist:

- [ ] Tests provide adequate coverage
- [ ] Code is well-commented
- [ ] New features are translatable or updates will not break translations
- [ ] Relevant documentation has been added or updated
- [ ] User impact is well-understood and desirable
- [ ] Pull Request is labeled appropriately
- [ ] Follow-up work items (including potential tech debt) are tracked and linked

PR #212, opened 2019-12-13
Use /docs instead of /documentation for links

dmcavoy:
# Description

Continuing of #210 and #211.

This changes localhost, curriculum.code.org, and codecurricula.com to user /docs instead of /documentation. /documentation will still work tho if someone has an old link to it. This is one of two steps in getting all the urls for documentation closer to using the same paths.

Next will be figuring out how to make docs.code.org use the same pattern as well.

## Links

- [jira](https://codedotorg.atlassian.net/browse/LP-910)

## Testing story

Still haven't figured out the best way to test this all yet.

# Reviewer Checklist:

- [ ] Code is well-commented

### DIFF
--- a/documentation/templatetags/documentation_extras.py
+++ b/documentation/templatetags/documentation_extras.py
@@ -8,12 +8,8 @@ register = template.Library()
 
 @register.filter
 def get_absolute_url_for_host(map, host):
-    if host == 'curriculum.code.org' or host == 'www.codecurricula.com' or host == 'localhost:8000' or host == 'testserver':
-        return '/documentation/%s/' % map.slug
-    elif host == 'studio.code.org':
+    if host == 'testserver' or host == 'localhost:8000':
         return '/docs/%s/' % map.slug
-    elif host == 'docs.code.org':
-        return '/%s/' % map.slug
     else:
         logger.info("no known host %s" % host)
-        return '/%s/' % map.slug
+        return '/docs/%s/' % map.slug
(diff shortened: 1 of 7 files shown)


documentation/templatetags/documentation_extras.py
```diff
@@ -8,12 +8,8 @@
 
 @register.filter
 def get_absolute_url_for_host(map, host):
```
dmcavoy: Once we verify that this is working I will just simply this to a method in the Map Model but for now, I wanted to stick with what we knew.

documentation/templatetags/documentation_extras.py
```diff
@@ -8,12 +8,8 @@
...
-    elif host == 'studio.code.org':
+    if host == 'testserver' or host == 'localhost:8000':
         return '/docs/%s/' % map.slug
-    elif host == 'docs.code.org':
-        return '/%s/' % map.slug
     else:
         logger.info("no known host %s" % host)
-        return '/%s/' % map.slug
\ No newline at end of file
+        return '/docs/%s/' % map.slug
```
davidsbailey: since this URL pattern matches the above, there should be no need for this conditional any more

documentation/templatetags/documentation_extras.py
davidsbailey: sorry, thanks for explaining, I think that means you can ignore my other comment below.

PR #218, opened 2019-12-18
Show Icon Next to Content Associated With

dmcavoy:
# Description

The link icons for Teaching Tips (and other things of that style) have been showing up in the left hand column but not always next to the content they are associated with. This change moves them to be inline with the content they are associated with instead of in the left hand column (see example below). We moved them inline because there is no easy way to know how many icons we want to support in the left hand column next to one thing at a time so this solution seems more scalable. 

Before this gets merged we will need to alert the curriculum team as this will mean doing a pass of lessons before they publish them. 

### Markdown

```
tip!!!tip-0 discussion!!!discussion-0 slide!!!slide-0 **Prompt:** The first app we are going to investigate uses a loops in a simulation. What is a simulation? Why are they useful?
```

### Result
Screen Shot 2020-01-10 at 8 55 36 AM

## Links

- [jira](https://codedotorg.atlassian.net/browse/LP-908)

## Testing story

Welcome suggestions for best way to approach testing and/or worth testing.

# Reviewer Checklist:

- [ ] Code is well-commented


lessons/static/css/commonlesson.css
```diff
@@ -383,12 +383,10 @@ html, body{
 }
 
 .tiplink {
-    float: left;
     width: 20px;
     font-size: 15px;
-    margin-left: -20px;
     clear: left;
-    display: block;
+    display: inline;
```
Hamms: we could probably get rid of this property altogether, as spans default to displaying inline

PR #226, opened 2020-01-07
Rendering test for standards page and standards on lesson page

dmcavoy:
# Description

Test that the standards page for a course and a lesson plan with standards render correctly.

Also creates basic factories for standards models.

## Links

- [jira](https://codedotorg.atlassian.net/browse/LP-1052)

# Reviewer Checklist:

- [ ] Code is well-commented


standards/factories.py
```diff
@@ -0,0 +1,34 @@
...
+    name = Sequence(lambda n: "GradeBand %03d" % n)
+
+class CategoryFactory(DjangoModelFactory):
+    class Meta:
+        model = Category
+
+    name = Sequence(lambda n: "Category %03d" % n)
+    framework = SubFactory('standards.factories.FrameworkFactory')
+    # Need to set parent to none on subfactory so that we don't end up with infinite recursion
+    parent = SubFactory('standards.factories.CategoryFactory', parent=None)
```
davidsbailey: nice tactic. not sure if we'll always need a category to have a parent, but if we find that we're making lots of CategoryFactory calls with parent=null then we could always refactor this later to have no parent be the default.

PR #230, opened 2020-01-07
Add standards to the serializer so they show up in unit meta data

Erin007:
We need to make the standards associated with lessons from Curriculum Builder available to Code Studio in order to group lessons by standard in the progress tab of the teacher dashboard ([spec](https://docs.google.com/document/d/1laOWuBRZYE9Xe8rxhaFyfJNXZwgukc8Hy7MAnKFl-Ws/edit?ts=5e14f489),[ implementation plan](https://docs.google.com/document/d/1laOWuBRZYE9Xe8rxhaFyfJNXZwgukc8Hy7MAnKFl-Ws/edit?ts=5e14f489)). 

Here, as a proof of concept, I added standards information to a new standards metadata endpoint similar to other metadata endpoints used to pass stage descriptions from Curriculum Builder -> Code Studio. 

Which results in: 
Screen Shot 2020-01-07 at 4 33 42 PM

We can then use this data in upcoming work to populate the standards view of the progress tab: 

Screen Shot 2020-01-07 at 2 37 54 PM


curricula/views.py
```diff
@@ -1074,6 +1074,16 @@ def stage_element(request, stage, format=None):
     serializer = UnitSerializer(unit)
     return Response(serializer.data)
 
+@api_view(['GET', ])
+def stage_standards(request, stage, format=None):
+    try:
+        unit = Unit.objects.get(login_required=False, status=2, stage_name=stage)
+    except MultipleObjectsReturned:
+        logger.exception("Warning - found multiple units referencing the stage %s" % stage)
```
uponthesun: Can this case realistically happen? If so, what would it mean?

In general I think it's better to raise exceptions rather than swallowing if its actually an unexpected case.

davidsbailey: good question. yes, it happens frequently: https://codedotorg.slack.com/archives/C0SUN2VT5/p1578388154147900  

It seems problematic to me too, but has also been declared by Josh C to not be worth doing anything about. Not sure if this is worth tracking at the moment but I didn't comment on this since it seems better to follow the existing pattern, and if we decide to change it we can change it everywhere.

dmcavoy: This only happens frequently for Professional Learning Courses as far as I know and thats because there is not actually a code studio course associated with the CB course. So we just make up something fake to go in the field and when a course gets cloned it usually doesn't get updated.

davidsbailey: To add a bit more background, because we're not quite sure how all the platformization stuff will shake out, we don't really know what the future of Curriculum Builder will be just yet -- which makes me want to push pause on this granularity of cleanup task for the moment. If we do decide to double down on CB, then I think it would start making sense to track stuff like this, which we'll notice because it is spamming a slack room which we monitor (so it may not be totally fair to say we are swallowing this error).

davidsbailey: > This only happens frequently for Professional Learning Courses as far as I know and thats because there is not actually a code studio course associated with the CB course. So we just make up something fake to go in the field and when a course gets cloned it usually doesn't get updated.

thank you for that background @dmcavoy . It sounds like all the noise from the PLC errors would probably hide a problem if a regular code studio course had this problem, so maybe it is a bit worse than I described. I could go either way on whether this is worth tracking then. If anyone feels inclined, it could go in the PLAT project in Jira.

dmcavoy: I'm still not sure its worth tracking right now.

standards/tests/test_standards_rendering.py
```diff
@@ -31,3 +31,15 @@ def test_render_lesson_includes_standard(self):
         # Standards Alignment header only shows if there are standards
         self.assertIn('Standards Alignment', response.content)
         self.assertIn('Test Standard', response.content)
+
+    def test_metadata_for_course(self):
```
Erin007: I don't know if the first two tests I wrote actually belong here since they aren't related to standards, but I didn't know where else to put them and it seemed better to have them than to not. @dmcavoy what do you think?

dmcavoy: Might make a little more sense for the course and unit one to live here: https://github.com/code-dot-org/curriculumbuilder/blob/master/curricula/tests/test_curricula_rendering.py

PR #233, opened 2020-01-08
Fix Map Menu Links - Move toward /docs a primary

dmcavoy:
# Description

## Problem 1: Too many routes to Maps
There are three was to look at Maps right now.

```
curriculum.code.org/documentation/concepts
docs.code.org/concepts
studio.code.org/docs/concepts
```

After going through and moving all current references to `studio.code.org/docs` this is follow up to make it possible to go to `curriculum.code.org/docs/concepts` in addition to `curriculum.code.org/documentation`.

## Problem 2: Map Menu Links

The map menu links had been just pointing to `/concepts`. This updates them to point to `/docs/concepts` so that they will work. With the changes to part 1 they should definitely work on `curriculum.code.org` and `studio.code.org`. 

## Links

- [jira](https://codedotorg.atlassian.net/browse/LP-910)

## Testing story

I updated the rendering tests for documentation to check that /documentation and /docs will both work for documentation. 

# Reviewer Checklist:

- [ ] Code is well-commented


curricula/urls.py
```diff
@@ -9,6 +9,7 @@
 
                        # Allow viewing concept documentations without the /documentation/ prefix
                        # so links work on both curriculumbuilder and docs.code.org
+                       url(r'^(?P<slug>/concepts.*)/$', documentation_views.map_view, name="map_view"),
```
dmcavoy: This adds in the old code that used to be here. The line below was changed in the process of trying to figure out how to make these links but never got reverted.

dmcavoy: I thought having both would be good in hopes that it would catch old urls that pointed to this page but hopefully the /docs/concepts route will work to move everything over to using the same url format

davidsbailey:
> There was an old change to curricula/urls.py that I think should have made it possible to go to docs.code.org/docs/concepts but it doesn't seem to work right now. I'm adding back in the old line that was in that file so both are an option. I wonder if we need to publish for docs.code.org to get the change?

I believe docs.code.org is configured in CloudFront to point to s3://cdo-curriculum/documentation, so I am suspicious of the statement that and old change to curricula/urls.py could affect its behavior. Please ping me if you would like to chat more about this.

curricula/urls.py
```diff
@@ -9,7 +9,7 @@
 
                        # Allow viewing concept documentations without the /documentation/ prefix
                        # so links work on both curriculumbuilder and docs.code.org
-                       url(r'^(?P<slug>/docs/concepts.*)/$', documentation_views.map_view, name="map_view"),
```
dmcavoy: Resetting this back. Was first changed by me here: https://github.com/code-dot-org/curriculumbuilder/commit/d527537eaf84ebc2c309fb9a57d60c3014c267c6 in attempt to change this 2 months ago.

documentation/admin.py
```diff
@@ -42,9 +42,9 @@ class IDEAdmin(PageAdmin, VersionAdmin):
         }),
     )
 
-    # Override the "view on site" button to use local /documentation/ instead of docs.code.org path
+    # Override the "view on site" button to use local /docs/ instead of slug which does not have /docs
     def view_on_site(self, obj):
-        return '/documentation%s' % obj.get_absolute_url()
+        return '/docs%s' % obj.get_absolute_url()
```
davidsbailey: It's a bit misleading that the method called `get_absolute_url` is not returning an absolute url. It seems like we still need that method though, because we are using its return value as a suffix after /docs/ as well as /documentation/ . Could you rename this to `get_url_suffix` or `get_url_part` or something like that?

If you are planning to change the rest of the urls over to use /docs instead of /documentation, then another idea would be to wait until that is done and then have `get_absolute_url` actually return an absolute url, including /docs/ prefix.

davidsbailey: oops, just realized this is overriding a django method: https://docs.djangoproject.com/en/1.8/ref/models/instances/#get-absolute-url

I'm not sure when this method is getting used by django behind the scenes, but it makes me think that it's important to define it accurately, so that it produces a working url, including /docs. then you can change the implementations of the methods which append this to '/documentation' to build their url a different way.

dmcavoy: I think I'm going to hold on this for now until we figure out docs.code.org - then we can just change the slug for everything over to /docs/..... and do the clean up work then. I will add it to my list of things to follow up on in jira

PR #241, opened 2020-01-16
Limit pdf generation to a handful of languages

bethanyaconnor:
# Description

We've wanted to add more language options to CB for a while. We haven't as the sync takes several hours every day. We believe that PDF generation is a major bottleneck, so this PR adds a way to limit the languages that generate translated PDFs. Languages that don't have translated PDFs will link to the English version.

This PR does add Hindi as a language option as I was testing these features out with Hindi. If we would prefer to keep adding a language separate, I can remove it.

# Reviewer Checklist:

- [ ] Code is well-commented

### DIFF
--- a/curricula/models.py
+++ b/curricula/models.py
@@ -10,6 +10,7 @@ from django.db.models import Count, Q
 from django.db.models.signals import post_save
 from django.dispatch import receiver
 from django.utils.text import slugify
+from django.utils.translation import get_language
 
 from mezzanine.pages.models import Page, RichText, Orderable, PageMoveException
 from mezzanine.core.fields import RichTextField
@@ -104,7 +105,12 @@ class Curriculum(InternationalizablePage, RichText, CloneableMixin, Ownable):
         return reverse('curriculum:curriculum_view', args=[self.slug])
 
     def get_pdf_url(self):
-        return reverse('curriculum:curriculum_pdf', args=[self.slug])
+        url = reverse('curriculum:curriculum_pdf', args=[self.slug])
+        language = get_language()
+        if (language != 'en-us' and language not in settings.LANGUAGE_GENERATE_PDF):
+            m = re.match(r'(/[^/]*)(/.*$)', url)
+            url = m.groups()[1]
+        return url
 
     def get_json_url(self):
         return reverse('curriculum:curriculum_element', args=[self.slug])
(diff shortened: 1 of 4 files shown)


curricula/models.py
```diff
@@ -104,7 +105,12 @@ def get_absolute_url(self):
         return reverse('curriculum:curriculum_view', args=[self.slug])
 
     def get_pdf_url(self):
-        return reverse('curriculum:curriculum_pdf', args=[self.slug])
+        url = reverse('curriculum:curriculum_pdf', args=[self.slug])
+        language = get_language()
+        if (language != 'en-us' and language not in settings.LANGUAGE_GENERATE_PDF):
+            m = re.match(r'(/[^/]*)(/.*$)', url)
+            url = m.groups()[1]
+        return url
```
Hamms: would it be feasible to instead do something like

```python
def get_pdf_url(self):
  language = get_language()
  if language != settings.LANGUAGE_CODE and language not in settings.LANGUAGE_GENERATE_PDF:
    language = settings.LANGUAGE_CODE
  with translation.override(language):
    return reverse('curriculum:curriculum_pdf', args=[self.slug])
```

i18n/management/commands/publish_i18n.py
```diff
@@ -60,7 +65,7 @@ def publish_models(self):
                 success_count, total, name, datetime.timedelta(seconds=int(elapsed_time))
             ))
 
-        log("Publishing %s models in %s languages finished in %s (average of ~%s per language" % (
+        log("Publishing %s models in %s languages finished in %s (average of ~%s per language)" % (
```
Hamms: typo?

i18n/management/commands/publish_i18n.py
```diff
@@ -19,7 +19,8 @@ class Command(BaseCommand):
     ]
 
     language_codes = [
-        language_code for language_code, _ in settings.LANGUAGES
+        (language_code, language_code in settings.LANGUAGE_GENERATE_PDF)
```
Hamms: is there a specific reason to add this here instead of inferring it explicitly below? I worry about a loss of readability

i18n/management/commands/publish_i18n.py
bethanyaconnor: Fixing the typo 
Screenshot 2020-01-17 at 9 08 12 AM

curricula/tests/test_curricula_rendering.py
```diff
@@ -172,3 +173,10 @@ def test_metadata_for_course(self):
     def test_metadata_for_unit(self):
         response = self.client.get('/metadata/test-stage-name.json')
         self.assertEqual(response.status_code, 200)
+
+    def test_get_pdf_url(self):
+        en_url = self.csf_unit.get_pdf_url()
+        with translation.override('es-mx'):
+            es_url = self.csf_unit.get_pdf_url()
+        with translation.override('hi-in'):
+            hi_url = self.csf_unit.get_pdf_url()
```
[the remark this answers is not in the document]
bethanyaconnor: I was having some merge conflict issues so I guess they disappeared...

curricula/models.py
bethanyaconnor: oh that's much better!

i18n/management/commands/publish_i18n.py
bethanyaconnor: Changed

i18n/management/commands/publish_i18n.py
Hamms: oooh, i totally misread this as the paren being outside the quote

PR #242, opened 2020-01-21
Change stage_name to unit_name to avoid further confusion

Erin007:
Units in CurriculumBuilder align with Scripts in Code Studio. For example 'coursea-2019' references a Unit in Curriculum Builder and a Script in Code Studio. 

Lessons in CurriculumBuilder align with Stages in Code Studio, for example "Learn to Drag and Drop". 

While sending standards metadata from CurriculumBuilder to Code Studio, it was noted that the  `stage_name` attribute on CurriculumBuilder Units was erroneously labeled. ([see discussion on Code Studio PR](https://github.com/code-dot-org/code-dot-org/pull/32743#discussion_r367729866)).

This updates `stage_name` to `unit_name` and changes the Unit model to have a `unit_name` attribute that references the `stage_name` column in the database. The end result is that the standards metadata now accurately references `unit_name`.

unit_name

### DIFF
--- a/curricula/models.py
+++ b/curricula/models.py
@@ -298,7 +298,7 @@ class Unit(InternationalizablePage, RichText, CloneableMixin, Ownable):
     ancestor = models.ForeignKey('self', blank=True, null=True, on_delete=models.SET_NULL)
     disable_numbering = models.BooleanField(default=False, help_text="Override to disable unit numbering")
     number = models.IntegerField('Number', blank=True, null=True)
-    stage_name = models.CharField('Script', max_length=255, blank=True, null=True,
+    unit_name = models.CharField('Script', max_length=255, blank=True, db_column='stage_name', null=True,
                                   help_text='Name of Code Studio script')
     questions = RichTextField('Support Details', help_text='Open questions or comments to add to all lessons in unit',
                               blank=True, null=True)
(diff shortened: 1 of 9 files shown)


Erin007:
I updated the tests specific to metadata which were previously referencing `stage_name` but this could potentially be a widespread scary change since the re-naming applies in multiple places.

curricula/views.py
```diff
@@ -1066,21 +1066,21 @@ def unit_element(request, slug, unit_slug, format=None):
 @api_view(['GET', ])
 def stage_element(request, stage, format=None):
     try:
-        unit = Unit.objects.get(login_required=False, status=2, stage_name=stage)
+        unit = Unit.objects.get(login_required=False, status=2, unit_name=stage)
     except MultipleObjectsReturned:
         logger.exception("Warning - found multiple units referencing the stage %s" % stage)
-        unit = Unit.objects.filter(login_required=False, status=2, stage_name=stage).first()
+        unit = Unit.objects.filter(login_required=False, status=2, unit_name=stage).first()
```
davidsbailey: please rename `stage` to `unit_name` in this method's code and error text

curricula/models.py
```diff
@@ -298,7 +298,7 @@ class Unit(InternationalizablePage, RichText, CloneableMixin, Ownable):
     ancestor = models.ForeignKey('self', blank=True, null=True, on_delete=models.SET_NULL)
     disable_numbering = models.BooleanField(default=False, help_text="Override to disable unit numbering")
     number = models.IntegerField('Number', blank=True, null=True)
-    stage_name = models.CharField('Script', max_length=255, blank=True, null=True,
+    unit_name = models.CharField('Script', max_length=255, blank=True, db_column='stage_name', null=True,
```
davidsbailey: nice find @Erin007 ! This seems like a great option.

PR #246, opened 2020-01-30
Add nine more languages to CB (none will have PDFs generated)

bethanyaconnor:
[FND-982](https://codedotorg.atlassian.net/browse/FND-982) lists 30 languages to add to the CB sync -- this PR covers the first 10 (Italian is listed on the ticket which is already synced).

I'm breaking this up so that we can monitor for any errors and fix them before enabling more languages.

### Time added
I ran the `publish_i18n` step with PDF generation and without with 6 languages added.

Without PDF generation:
```Publishing 3 models in 6 languages finished in 0:21:07 (average of ~0:03:31 per language)```

With PDF generation on 4 languages:
```Publishing 3 models in 6 languages finished in 3:12:46 (average of ~0:32:07 per language)```

When I ran the sync with out PDF generation with all the newly added languages:
```Publishing 3 models in 15 languages finished in 0:45:49 (average of ~0:03:03 per language)```

These number suggest that each new language _without PDF generation_ will add 3-4 minutes to 
the publish step (there will likely be some added time in the down step as well). This seems like a tolerable amount of time.

## Testing story

Ran the down and out steps locally and tested the site on a handful of languages. No issues.

# Reviewer Checklist:

- [ ] Code is well-commented

### DIFF
--- a/curriculumBuilder/settings.py
+++ b/curriculumBuilder/settings.py
@@ -127,20 +127,29 @@ LANGUAGE_CODE_DO_TRANSLATION = "in-tl"
 
 # Supported languages
 LANGUAGES = (
+    ('ar-sa', _('Arabic')),
+    ('de-de', _('German')),
     ('en-us', _('English')),
     ('es-mx', _('Mexican Spanish')),
+    ('fr-fr', _('French')),
+    ('hi-in', _('Hindi')),
     ('it-it', _('Italian')),
-    ('th-th', _('Thai')),
+    ('ja-jp', _('Japanese')),
+    ('ko-kr', _('Korean')),
+    ('ru-ru', _('Russian')),
     ('sk-sk', _('Slovak')),
+    ('th-th', _('Thai')),
+    ('tr-tr', _('Turkish')),
+    ('zh-cn', _('Simplified Chinese')),
+    ('zh-tw', _('Traditional Chinese')),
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: curriculumBuilder/settings.py [hunk 1]
```diff
@@ -127,20 +127,29 @@
...
 # Supported languages
 LANGUAGES = (
+    ('ar-sa', _('Arabic')),
+    ('de-de', _('German')),
     ('en-us', _('English')),
     ('es-mx', _('Mexican Spanish')),
+    ('fr-fr', _('French')),
+    ('hi-in', _('Hindi')),
     ('it-it', _('Italian')),
-    ('th-th', _('Thai')),
```
bethanyaconnor: I arranged the languages into alphabetical order by language code, which is why there are "deleted" lines

daynew:
Should `With PDF generation on 4 languages:` and `Without PDF generation:` be swapped in this PR description?

dju90:
This test seems to indicate that PDF generation is the only major bottleneck in adding languages to the CB sync. Would that be a correct interpretation?

bethanyaconnor:
@daynew you're right 🤦‍♀ Fixed.

@dju90 yep that's my conclusion! The only difference between the two runs was adding `False and` to where we check if we should publish a PDF (i.e. making it so we never publish the PDF). Strongly suggests the PDF generation takes about 3 hours. :)

PR #253, opened 2020-02-11
Fix the links on the docs landing page

dmcavoy:
The links on studio.code.org/docs are not working because they need the trailing slash at the end to work. This makes that update so that we can generate that page to update it on S3.

### DIFF
--- a/documentation/models.py
+++ b/documentation/models.py
@@ -35,6 +35,9 @@ class IDE(Page, RichText, CloneableMixin):
     def get_absolute_url(self):
         return '/%s/' % self.slug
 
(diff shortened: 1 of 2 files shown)


clareconstantine:
Are trailing slashes a django requirement? I did some googling about trailing slashes in urls, but I'm curious at what level it's being enforced in this case.

dmcavoy:
The trailing slash issue seems to only be something on studio.code.org/docs/ and I think it might be because of the curriculum proxy and or the path we use. I need to look into it because it keeps being an issue.

PR #256, opened 2020-02-14
Add standards metadata to the list of urls that update when a unit is published 

Erin007:


uponthesun:
Code LGTM - have you tested it locally?

davidsbailey:
@Erin007 I need to do something similar to this . were you able to test it locally?

Erin007:
> @Erin007 I need to do something similar to this . were you able to test it locally?

I don't think so.  If I remember correctly, I just merged this PR, re-published the units and then had the standards data we need available for import.

PR #262, opened 2020-03-02
Catch and log pdf errors in publish_i18n without halting the whole pipeline

bethanyaconnor:
# Description

There are sometimes pdf publishing issues that are a result of content changes. While we should fix those, I don't think the whole pipeline should halt while we do.

## Links

- [spec]()
- [jira]()

## Testing story

# Reviewer Checklist:

- [ ] Code is well-commented

### DIFF
--- a/i18n/management/commands/publish_i18n.py
+++ b/i18n/management/commands/publish_i18n.py
@@ -56,7 +56,8 @@ class Command(BaseCommand):
                           list(obj.publish_pdfs(silent=True))
                           pdf_generation_end_time = time.time()
                           total_pdf_generation_time += (pdf_generation_end_time - pdf_generation_start_time)
-                        except:
+                        except Exception as e:
+                          log(e)
                           log("PDF publishing failed %s in %s" % (obj.slug, language_code))
                 success_count += 1
 
--- a/i18n/management/commands/publish_i18n.py
+++ b/i18n/management/commands/publish_i18n.py
@@ -51,10 +51,13 @@ class Command(BaseCommand):
                         list(obj.publish(silent=True))
                     # Temporary hack to turn off PDF generation while still directing users to the translated pdfs if they exist
                     if language_code in settings.LANGUAGE_GENERATE_PDF and hasattr(obj, 'publish_pdfs'):
-                        pdf_generation_start_time = time.time()
-                        list(obj.publish_pdfs(silent=True))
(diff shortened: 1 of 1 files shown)


i18n/management/commands/publish_i18n.py
```diff
@@ -51,10 +51,13 @@ def publish_models(self):
...
-                        list(obj.publish_pdfs(silent=True))
-                        pdf_generation_end_time = time.time()
-                        total_pdf_generation_time += (pdf_generation_end_time - pdf_generation_start_time)
+                        try:
+                          pdf_generation_start_time = time.time()
+                          list(obj.publish_pdfs(silent=True))
+                          pdf_generation_end_time = time.time()
+                          total_pdf_generation_time += (pdf_generation_end_time - pdf_generation_start_time)
+                        except:
+                          log("PDF publishing failed %s in %s" % (obj.slug, language_code))
```
mvkski: Could we also log the exception stacktrace and error message, so that we could root cause what the speicfic issues with PDF generation are?

i18n/management/commands/publish_i18n.py
```diff
@@ -51,10 +51,13 @@ def publish_models(self):
                         list(obj.publish(silent=True))
                     # Temporary hack to turn off PDF generation while still directing users to the translated pdfs if they exist
                     if language_code in settings.LANGUAGE_GENERATE_PDF and hasattr(obj, 'publish_pdfs'):
-                        pdf_generation_start_time = time.time()
-                        list(obj.publish_pdfs(silent=True))
-                        pdf_generation_end_time = time.time()
-                        total_pdf_generation_time += (pdf_generation_end_time - pdf_generation_start_time)
+                        try:
+                          pdf_generation_start_time = time.time()
+                          list(obj.publish_pdfs(silent=True))
```
Hamms: any reason not to put this catch in the publish_pdfs method itself?

i18n/management/commands/publish_i18n.py
bethanyaconnor: Logged the exception

i18n/management/commands/publish_i18n.py
bethanyaconnor: It doesn't seem unreasonable for the page to throw an error if the pdf is failing to generate

PR #265, opened 2020-03-03
Fixpdfcb

mvkski:
# Description

This should fix PDF generation failing for curriculumbuilder. In this change, we migrate off 
of wkhtmltopdf and onto puppeteer, a library from Google that Winter researched. Using this library,
pdf generation succeeds on pages where it previously failed. 

## Links

- [spec]()
- [jira]()

## Testing story

Test manually. As is, it is difficult to test this feature across the entire site
in an automated way. We will be able to do this easily if we write a PDF generation service.

# Reviewer Checklist:

- [ ] Code is well-commented

### DIFF
--- a/curriculumBuilder/settings.py
+++ b/curriculumBuilder/settings.py
@@ -534,7 +534,8 @@ PHANTOMJS_KEY = os.environ.get('PHANTOMJS_KEY')
 AWS_QUERYSTRING_AUTH = False
 AWS_ACCESS_KEY_ID = os.environ.get('AWS_ACCESS_KEY_ID')
 AWS_SECRET_ACCESS_KEY = os.environ.get('AWS_SECRET_ACCESS_KEY')
-AWS_STORAGE_BUCKET_NAME = 'cdo-curriculum'
+AWS_SESSION_TOKEN = os.environ.get('AWS_SESSION_TOKEN')
+AWS_STORAGE_BUCKET_NAME = os.environ.get('AWS_STORAGE_BUCKET_NAME') or 'cdo-curriculum'
 # AWS_S3_CUSTOM_DOMAIN = '%s.s3.amazonaws.com' % AWS_STORAGE_BUCKET_NAME
 AWS_S3_CUSTOM_DOMAIN = 'curriculum.code.org'
 AWS_PRELOAD_METADATA = True  # helps collectstatic do updates
@@ -545,11 +546,11 @@ AWS_HEADERS = {
 AWS_BASE_URL = 'http://cdo-curriculum.s3-website-us-east-1.amazonaws.com'
 
 STATICFILES_LOCATION = 'static'
-STATICFILES_STORAGE = 'curriculumBuilder.s3utils.StaticRootS3BotoStorage'
+STATICFILES_STORAGE = 'helpers.s3utils.StaticRootS3BotoStorage'
 STATIC_URL = "https://%s/%s/" % (AWS_S3_CUSTOM_DOMAIN, STATICFILES_LOCATION)
 
 MEDIAFILES_LOCATION = 'media'
-DEFAULT_FILE_STORAGE = 'curriculumBuilder.s3utils.MediaRootS3BotoStorage'
+DEFAULT_FILE_STORAGE = 'helpers.s3utils.MediaRootS3BotoStorage'
 MEDIA_URL = "https://%s/%s/" % (AWS_S3_CUSTOM_DOMAIN, MEDIAFILES_LOCATION)
 
 I18N_STORAGE = 'storages.backends.s3boto.S3BotoStorage'
@@ -585,7 +586,7 @@ FREEZE_INCLUDE_STATIC = False
 # operation which is used for every publish. Disabling preloading here should
 # speed up publishing content by about an order of magnitude.  See
 # https://stackoverflow.com/a/21121924/1810460 for more details.
-JACKFROST_STORAGE = 'storages.backends.s3boto.S3BotoStorage'
+JACKFROST_STORAGE = 'helpers.s3utils.S3Boto3StorageSTS'
 JACKFROST_STORAGE_KWARGS = {
     'preload_metadata': False
 }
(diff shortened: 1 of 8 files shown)


helpers/pdfUtils.py
```diff
@@ -0,0 +1,20 @@
...
+
+SCRIPT_PATH = settings.BASE_DIR + "/bin/generate-pdf.js" 
+
+def get_pdf_for_url(url):
+  try:
+    tmp = tempfile.NamedTemporaryFile(delete=True)
+    child = subprocess.Popen(["node", SCRIPT_PATH, "-w", tmp.name, "-u %s" % (url)], stdout=subprocess.PIPE)
+    child.communicate()
+    contents = tmp.read()
+    tmp.close()
```
uponthesun: Should use a python `with` block instead - it's quite nice! Python's version of Java try with resources (though I think it came earlier?)

helpers/pdfUtils.py
```diff
@@ -0,0 +1,20 @@
...
+from subprocess import CalledProcessError
+
+logger = logging.getLogger(__name__)
+
+SCRIPT_PATH = settings.BASE_DIR + "/bin/generate-pdf.js" 
+
+def get_pdf_for_url(url):
+  try:
+    tmp = tempfile.NamedTemporaryFile(delete=True)
+    child = subprocess.Popen(["node", SCRIPT_PATH, "-w", tmp.name, "-u %s" % (url)], stdout=subprocess.PIPE)
```
uponthesun: Docs suggest `subprocess.call` is simplest for common uses, are there specific reasons we need to use `Popen`? https://docs.python.org/2/library/subprocess.html

helpers/s3utils.py
```diff
@@ -60,3 +61,21 @@ def save(self, name, content):
...
+#
+# see:
+#    https://github.com/jschneier/django-storages/issues/282
+#    https://github.com/jschneier/django-storages/blob/9f07eab9c2f86c3d911ec9f6b0f45dca36626bd1/storages/backends/s3boto3.py#L242
+#
+# TODO: When we upgrade django to a version past 1.11, we will be able to
+# upgrade django-storages to a version that does not have this bug, and delete
+# the below class
+class S3Boto3StorageSTS(S3Boto3Storage):
+  access_key = False
```
uponthesun: Have you tested to make sure this works with just access_key and secret_key as well, like in production?

curriculumBuilder/settings.py
```diff
@@ -545,11 +546,11 @@ def iframe_filter_domain(name, value):
 AWS_BASE_URL = 'http://cdo-curriculum.s3-website-us-east-1.amazonaws.com'
 
 STATICFILES_LOCATION = 'static'
-STATICFILES_STORAGE = 'curriculumBuilder.s3utils.StaticRootS3BotoStorage'
+STATICFILES_STORAGE = 'helpers.s3utils.StaticRootS3BotoStorage'
```
uponthesun: Why the file move? not opposed to it, just wondering.

helpers/pdfUtils.py
mvkski: Its a holdover from a previous version of this function where the PDF contents were read from STDOUT. Will switch to subprocess.call

helpers/s3utils.py
mvkski: I haven't tested it, but it should work reading the source for S3Boto3Storage:
https://github.com/jschneier/django-storages/blob/9f07eab9c2f86c3d911ec9f6b0f45dca36626bd1/storages/backends/s3boto3.py#L242
Setting the class variables to false doesn't actually keep the credentials from being loaded from environment variables.

I'd be happy to test locally if you know of a way to get the credentials onto my machine

curriculumBuilder/settings.py
mvkski: Arguably arbitrary personal preference. I thought it'd make sense to put pdfUtils into a 'helpers' module since it doesn't really seem to fit in anywhere. While I was at it, I figured I'd put the s3utils module in there too, since it also seems like a 'helper' and it bothered me that there was a file with the suffix utils in a module that isn't a utils module. 

I don't feel suuuper strongly about it being here vs somewhere else.

helpers/pdfUtils.py
mvkski: sounds good

helpers/s3utils.py
uponthesun: Sounds good then, we can just verify it after the deployment.

mvkski:
Good call on the ideal being splitting this up over 2 PRs. Will take note for the future

helpers/pdfUtils.py
```diff
@@ -10,11 +10,10 @@
...
-    tmp = tempfile.NamedTemporaryFile(delete=True)
-    child = subprocess.Popen(["node", SCRIPT_PATH, "-w", tmp.name, "-u %s" % (url)], stdout=subprocess.PIPE)
-    child.communicate()
-    contents = tmp.read()
-    tmp.close()
-    return contents
+    with tempfile.NamedTemporaryFile(delete=True) as tmp:
+      child = subprocess.call(["node", SCRIPT_PATH, "-w", tmp.name, "-u %s" % (url)])
+      contents = tmp.read()
+      tmp.close()
```
uponthesun: You don't need to call close, the `with` block handles that: http://effbot.org/zone/python-with-statement.htm

helpers/pdfUtils.py
```diff
@@ -10,11 +10,10 @@
...
+      return contents
   except CalledProcessError, e:
     logger.exception("getting pdf failed for url %s: %s" % (url, e.output))
```
uponthesun: Have you manually tested that the relevant error info gets logged?

helpers/pdfUtils.py
mvkski: right right, will remove

helpers/pdfUtils.py
mvkski: Yep! it does in fact get log

mvkski: CalledProcessError contains the output of the failed subprocess

PR #266, opened 2020-03-03
Use short_title in unit calendar if one is supplied

dmcavoy:
On the Unit Overview page there is a pacing calendar. For really long titles the text was overflowing the space. Josh had already added short_title as a field that was used in the PL session header which is very similar to the calendar on the unit overview page.

This just takes advantage of that field and uses the short_title in the unit calendar if one is supplied.

In the screenshots below look at the difference on Lesson 13.

### Before
Screen Shot 2020-03-03 at 2 37 17 PM

### After
Screen Shot 2020-03-03 at 2 36 12 PM


dmcavoy:
Yup the ones that are cutoff just don't have short titles yet. I just didn't feel like adding short titles to all those lessons.

PR #279, opened 2020-04-30
Override locale middleware

bethanyaconnor:
# Description

Django documentation a lot of this work was based on is [here](https://docs.djangoproject.com/en/1.8/topics/i18n/translation/#how-django-discovers-language-preference)

In the past, we've attempted to sync Polish but it caused problems because `/pl-sandbox/` resolved to be in Polish. It turns out that Django automatically uses the "base language" if it can't find the locale, so Django thought `sandbox` was a locale. This PR overrides that logic to only use the URL locale if it is an exact match to one we have enabled.

I rearranged the order of the middleware to be in line with the above documentation and because it wasn't working before I moved the `StrictLocaleMiddleware` to be after `SessionMiddleware`.

## Links

- [jira](https://codedotorg.atlassian.net/browse/FND-997)

## Testing story
- Added a unit test. 
- Manually tested `/pl-sandbox/csd_q1_and_q2/`, `/csf-1718/coursea/`, and `/es-mx/csf-1718/coursea/` to ensure they all still worked at intended

# Reviewer Checklist:

- [ ] Code is well-commented

### DIFF
--- a/i18n/middleware.py
+++ b/i18n/middleware.py
@@ -1,0 +1,36 @@
+import re
+import warnings
+
+from django.conf import settings
+from django.utils import translation
+from django.http import HttpResponseRedirect
+from django.middleware.locale import LocaleMiddleware
+
+def get_language_from_path_strict(path, allowed_languages):
+    language_code_prefix_re = re.compile(r'^/([\w-]+)(/|$)')
+    regex_match = language_code_prefix_re.match(path)
+    if not regex_match:
+        return settings.LANGUAGE_CODE
+    lang_code = regex_match.group(1)
+    if lang_code not in allowed_languages:
+        return settings.LANGUAGE_CODE
+    return lang_code
+
+class StrictLocaleMiddleware(LocaleMiddleware):
+
+    def __init__(self):
+        super(StrictLocaleMiddleware, self).__init__()
+        self.language_codes = [
+            language_code for language_code, _ in settings.LANGUAGES
+            if language_code != settings.LANGUAGE_CODE
+        ]
+
+    # See https://github.com/django/django/blob/6a0dc2176f4ebf907e124d433411e52bba39a28e/django/middleware/locale.py#L29
+    # for the original implementation
+    # This function simply checks for the language of the request by only looking at the URL.
+    # It also requires an exact language match in order to set the language (see above)
+    def process_request(self, request):
+        language = get_language_from_path_strict(
+                request.path_info, self.language_codes)
+        translation.activate(language)
+        request.LANGUAGE_CODE = translation.get_language()
--- a/i18n/middleware.py
+++ b/i18n/middleware.py
@@ -1,23 +1,22 @@
 import re
-import warnings
 
 from django.conf import settings
 from django.utils import translation
-from django.http import HttpResponseRedirect
 from django.middleware.locale import LocaleMiddleware
 
 def get_language_from_path_strict(path, allowed_languages):
-    language_code_prefix_re = re.compile(r'^/([\w-]+)(/|$)')
-    regex_match = language_code_prefix_re.match(path)
-    if not regex_match:
-        return settings.LANGUAGE_CODE
-    lang_code = regex_match.group(1)
-    if lang_code not in allowed_languages:
-        return settings.LANGUAGE_CODE
-    return lang_code
+    for language_code in allowed_languages:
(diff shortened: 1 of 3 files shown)


i18n/middleware.py
```diff
@@ -0,0 +1,36 @@
+import re
+import warnings
+
+from django.conf import settings
+from django.utils import translation
+from django.http import HttpResponseRedirect
+from django.middleware.locale import LocaleMiddleware
+
+def get_language_from_path_strict(path, allowed_languages):
+    language_code_prefix_re = re.compile(r'^/([\w-]+)(/|$)')
```
Hamms, line 10: is it necessary to use regex for this? I figure we could just do something like

```python
def get_language_from_path_strict(path, allowed_languages):
  for language_code in allowed_languages:
    if path.startswith('/{}/'.format(language_code)):
      return language_code

  return settings.LANGUAGE_CODE
```

i18n/tests/test_url_resolver.py
```diff
@@ -0,0 +1,36 @@
...
+from curricula.factories import UserFactory, CurriculumFactory, UnitFactory
+from lessons.factories import LessonFactory, ResourceFactory, ActivityFactory
+from django.conf import settings
+
+class I18nUrlResolverTestCase(TestCase):
+    def setUp(self):
+        self.csf_curriculum = CurriculumFactory(
+            slug="csf-curriculum",
+            unit_template_override='curricula/csf_unit.html')
+        self.pl_curriculum = CurriculumFactory(
```
Hamms, line 14: I'd love to see something explicitly explaining why we're testing with this curriculum in particular

i18n/middleware.py
```diff
@@ -0,0 +1,36 @@
+import re
+import warnings
+
+from django.conf import settings
+from django.utils import translation
+from django.http import HttpResponseRedirect
```
Hamms, line 6: is this used? or `warnings` above?

i18n/middleware.py
```diff
@@ -0,0 +1,36 @@
...
+        super(StrictLocaleMiddleware, self).__init__()
+        self.language_codes = [
+            language_code for language_code, _ in settings.LANGUAGES
+            if language_code != settings.LANGUAGE_CODE
+        ]
+
+    # See https://github.com/django/django/blob/6a0dc2176f4ebf907e124d433411e52bba39a28e/django/middleware/locale.py#L29
+    # for the original implementation
+    # This function simply checks for the language of the request by only looking at the URL.
+    # It also requires an exact language match in order to set the language (see above)
```
Hamms, line 31: this should probably go in the docstring for the class

i18n/middleware.py
bethanyaconnor: oh oops. It was at one point in my development but not anymore. I'll remove

i18n/middleware.py
bethanyaconnor: Changed

i18n/tests/test_url_resolver.py
```diff
@@ -0,0 +1,42 @@
...
+from django.contrib.auth.models import Permission
+from django.utils import translation
+
+from curricula.factories import UserFactory, CurriculumFactory, UnitFactory
+from lessons.factories import LessonFactory, ResourceFactory, ActivityFactory
+from django.conf import settings
+from mock import patch, Mock
+
+
+@patch('i18n.utils.I18nFileWrapper.get_translated_field', Mock())
```
Hamms, line 11: so clean!

i18n/tests/test_url_resolver.py
```diff
@@ -0,0 +1,42 @@
...
+    def test_render_curriculum(self):
+        response = self.client.get('/csf-curriculum/')
+        self.assertEqual(response.status_code, 200)
+        for language_code, _ in settings.LANGUAGES:
+            response = self.client.get('/%s/csf-curriculum/' % language_code)
+            self.assertEqual(response.status_code, 200, "failed for language %s" % language_code)
+            self.assertEqual(language_code, translation.get_language())
+        # Unfortunately, other tests assume that the language activated will be English
+        # so let's clean up after ourselves.
+        translation.activate(settings.LANGUAGE_CODE)
```
Hamms, line 37: this should probably go in a `tearDown` method, so we also do it for the other test case

PR #287, opened 2020-05-30
disable mdlightbox plugin

davidsbailey:
disable the mdlightbox plugin, which automatically wraps all images on curriculumbuilder in a [lightbox](https://en.wikipedia.org/wiki/Lightbox_(JavaScript)). there are two reasons for this:

1. As part of moving block and concept docs from Curriculum Builder to Level Builder, we need to eliminate html syntax from markdown. to accomplish that, we'd like to be able to use markdown to create images which are also links. The current lightbox plugin prevents that, because it overwrites the link target with a link to 

2. The current behavior is broken anyway (see screenshots). clicking the image navigates to the image, instead of opening a lightbox.

### before

Screen Shot 2020-05-30 at 10 34 48 AM

^ you can see at bottom left (and in the html) that the image is a simple link to itself
 
### after

Screen Shot 2020-05-30 at 10 22 20 AM

### DIFF
--- a/curriculumBuilder/settings.py
+++ b/curriculumBuilder/settings.py
@@ -483,7 +483,7 @@ PAGEDOWN_MARKDOWN_EXTENSIONS = (
     'curriculumBuilder.doclinks',
     'extra', 'codehilite', 'toc', 'curriculumBuilder.newtab', 'curriculumBuilder.absolute_images',
     # 'curriculumBuilder.resourcelinks', 'curriculumBuilder.highlightblocks', 'smarty',
-    'curriculumBuilder.resourcelinks', 'curriculumBuilder.mdlightbox',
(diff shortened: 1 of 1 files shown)


Hamms:
Is there a specific reason we're just disabling the plugin, rather than removing it and its dependencies entirely?

davidsbailey:
> Is there a specific reason we're just disabling the plugin, rather than removing it and its dependencies entirely?

I would plan to remove the plugin entirely in a follow-up PR, after waiting a short while to make sure this change didn't break anything too badly. If you think the risk is low, I could include it all in this PR.

davidsbailey:
I've implemented the rest of the cleanup in https://github.com/code-dot-org/curriculumbuilder/pull/288 . My plan would be to wait a couple days after shipping this PR before merging that PR.

PR #284, opened 2020-05-14
Embed app and code

mvkski:
This PR is the curriculumbuilder side of changes that allow us to embed projects directly in curriculumbuilder documentation. These changes add a new field,  'App Display Type', which specifies weather an app link in a documentation example should resolve to the new view (app with code from the project) or the old view (code is specified in curriculumbuilder, app is rendered from project). 

New view on docs page:
- [jira](https://codedotorg.atlassian.net/browse/LP-790)
Screenshot from 2020-05-14 00-08-40

New field in curriculum builder admin page:
Screenshot from 2020-05-14 00-14-19

## Links

# Reviewer Checklist:

- [ ] Tests provide adequate coverage
- [ ] Code is well-commented
- [ ] New features are translatable or updates will not break translations
- [ ] Relevant documentation has been added or updated
- [ ] User impact is well-understood and desirable
- [ ] Pull Request is labeled appropriately
- [ ] Follow-up work items (including potential tech debt) are tracked and linked

PR #291, opened 2020-06-17
Fix embed_app_with_code

mvkski:
There were some requests for additional work after the [original release](https://github.com/code-dot-org/curriculumbuilder/pull/284) of the feature. This PR contains

1. Fixing links to projects with special characters in the project name
2. Allowing for editors to specify the height of the embed_app_with_code view
3. Fix default value for app type
4. Fix template to show name and description for embed_app_with_code_view
5. add unit tests for the above

Detailed descriptions on how these changes relate to jira items can be found below

Here are how some of the changes look like. 
Before:
(window is stuck in small size, cannot scroll for examples with more code. Example name and description are missing)
Screenshot from 2020-05-14 00-08-40

After:
(height for iframe set to 1000px with new 'embed_app_with_code_height' field, example name and description is now rendered. Scrollbar allows user to scroll all the way to the bottom of code for example)

Screenshot from 2020-06-22 16-35-58

## Links

- [LP-1483 Project Examples Aren't scrollable](https://codedotorg.atlassian.net/browse/LP-1483)
Before this change, It wasnt possible to scroll to the bottom fo EMBED_APP_AND_CODE
examples that had a lot of code, because the size of the iframe for the example was 
too small. This change allows curriculum maintainers to resize the iframe for the example,which makes it possible to view all of the code for every example. When testing locally, I found that
if the iframe height is set to 1000px, examples of any size can be scrolled through.
- [LP-1485 Strange code showing up for some gamelab/applab docs](https://codedotorg.atlassian.net/browse/LP-1485)
This was due to a bug in the template for examples. The default value of the
app_display_type_field was set to the tuple (CODE_FROM_CODE_FIELD, 'Display app with code from code field above') but should have just been CODE_FROM_CODE_FIELD
- [LP-1475 On new Spritelab block docs, title/description texted entered on the admin page does not show](https://codedotorg.atlassian.net/browse/LP-1475)
The title/description fields were missing from the template for EMBED_APP_AND_CODE view,mbut have since been re-added
- [LP-1474 On new Spritelab docs, Project URLs with special characters like ‘-’ don’t work](https://codedotorg.atlassian.net/browse/LP-1474)
This bug was due to using the code that generates project links for the EMBED_APP_AND_CODE view using a regex that ignored special characters. The code was inherited from what was present in the project before we added spritelab support - spritelab projects can have special characters,
but the project types we supported did not so this wasn't caught until now. Fixed by using simpler code to generate these examples for EMBED_APP_AND_CODE view, without using a regex.
## Testing story

Added unit tests for new model features, verified changes visually

# Reviewer Checklist:

- [ ] Code is well-commented


documentation/migrations/0026_auto_20200616_0122.py
```diff
@@ -0,0 +1,19 @@
...
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('documentation', '0025_auto_20200616_0118'),
+    ]
+
+    operations = [
+        migrations.AlterField(
+            model_name='example',
+            name='embed_app_with_code_height',
```
davidsbailey, line 16: can you please consolidate these into a single migration? it looks like all 3 affect the same field.

documentation/models.py
```diff
@@ -299,11 +307,19 @@ def _append_suffix_to_app(self, suffix):
...
+    # this legacy code is kept to support data in the database that might expect it.
+    # ideally would use something closer to the code in 'get_embed_app_and_code'
+    # rather than the complicated regex munging below
     def get_embed_app(self):
-        return self._append_suffix_to_app('embed')
+         if self.app:
+            re_url = '\w*(studio.code.org\/p\w*\/\w+\/\w+)'
+            if re.search(re_url, self.app):
+                app_with_suffix = 'https://%s/%s' % (re.search(re_url, self.app).group(0), 'embed')
+                return app_with_suffix
```
davidsbailey, line 318: can this be deduplicated?
```suggestion
         return self._append_suffix_to_app('embed')
```

documentation/models.py
```diff
@@ -278,17 +279,24 @@ class Example(Orderable, CloneableMixin):
...
     app_display_type = models.CharField(
         max_length=255,
         choices=app_display_type_options,
-        default=app_display_type_options[0][0],
+        default=CODE_FROM_CODE_FIELD,
         help_text='How the app and code fields for this example are rendered')
 
+    embed_app_with_code_height = models.IntegerField(
+        'Embed app with code iframe height',
+        default=310,
```
davidsbailey, line 297: can you help me understand this default? the method name sounds super general, but then I'd be worried if this value was chosen super-specifically for spritelab, especially since it looks like it will take a DB migration to change it.

documentation/migrations/0026_auto_20200616_0122.py
mvkski: Sounds good, will do

documentation/models.py
mvkski: Not entirley sure what you mean by deduplicated here. I am going to hazard a guess that you're asking of we can combine the code for get_embed_app and get_embed_app_and_code.

I don't think this would be a good idea, because get_embed_app_and_code needs a different implementation to support special characters, but get_embed_app is supporting existing data in the database that I am concerned might cause bugs if we were to change it

documentation/models.py
mvkski: The values was chosen somewhat arbitrarily for what I thought looks good on spritelab, for examples that don't have a ton of code. My understanding is that at this time we are mostly expecting to use this feature for spritelab, but it would be nice to be able to use it for other features too.

Why is it a big concern that it would take a DB migration to change this field in the future? I am not currently aware of any big difficulties with doing migrations for curriculumbuilder - am I missing some context?

documentation/models.py
davidsbailey: I'm talking about calling `_append_suffix_to_app`. I made a github suggestion in my original comment to show it specifically.

documentation/models.py
davidsbailey: OK, this is probably fine.

clareconstantine:
@davidsbailey here is the PR for the original work, if that provides helpful context: https://github.com/code-dot-org/curriculumbuilder/pull/284

PR #294, opened 2020-07-01
Always Link Unit PDFs to curriculum.code.org

mvkski:
This is to ensure that Unit PDFs on codecurricula.com always show if they have
been published. Previously, the django app would re-generate Unit PDFs every
time they were requested, which could cause timeout errors that confuse site
visitors.

This change was made to address bug reports we got where folks requesting PDFs on codecurricula.com via these links would sometimes get error pages. They will now never get error pages if the PDFs for these pages have already been published - pdf publishing functionality has remained unchanged.

## Links

- [jira](https://codedotorg.atlassian.net/browse/LP-1495)

## Testing story
Tested by verifying links work as expected manually.

# Reviewer Checklist:

- [ ] Code is well-commented


clareconstantine:
I might be missing some context - what happens if the PDFs haven't been published?

mvkski:
If a PDF hasn't been published, this will 404. This is a good thing - if on codecurricula.com, this encourages an admin to publish the PDFs. If on https://curriculum.code.org/ , you'd get a 404 before this change as well.

mvkski:
PDFs were generated, but never published if this link was clicked. It has always been necessary to use the separate PDF publishing workflow accessible in the admin panel.

If a PDF hasn't been published, this will now 404. This is a good thing - if on codecurricula.com, this encourages an admin to publish the PDFs. If on https://curriculum.code.org/ , you'd get a 404 before this change as well.

PR #300, opened 2020-07-22
Parallelize "publish" operations in I18n sync

Hamms:
# Description

Specifically, parallelize the act of iterating through each model object. Testing on my local machine confirms that this speeds up the publish step proportional to the number of processor cores used. It looks like heroku dynos have between 2 and 4 cores, so we should expect a comparable speedup.

Once we've deployed this and looked into the results, I'd like to also explore the effects of allocating more resources to the dyno that runs the sync.

We could also look into parallelizing this at the language level rather than at the object level, but local testing doesn't indicate that that results in any significant improvement over this approach.

## Links

- [jira](https://codedotorg.atlassian.net/browse/FND-625)

## Testing story

Manually tested.

# Reviewer Checklist:

- [ ] Code is well-commented

### DIFF
--- a/i18n/management/commands/publish_i18n.py
+++ b/i18n/management/commands/publish_i18n.py
@@ -13,6 +13,31 @@ def can_publish_model(model):
     return hasattr(model, 'publish') or hasattr(model, 'publish_pdfs')
 
 
+def publish(obj):
+    pdf_generation_time = 0
+
+    if not obj.should_be_translated:
+        return pdf_generation_time
+
+    for language_code in Command.language_codes:
+        translation.activate(language_code)
+        if hasattr(obj, 'publish'):
+            list(obj.publish(silent=True))
+        # Temporary hack to turn off PDF generation while still directing users to the
+        # translated pdfs if they exist
+        if language_code in settings.LANGUAGE_GENERATE_PDF and hasattr(obj, 'publish_pdfs'):
+            try:
+                pdf_generation_start_time = time.time()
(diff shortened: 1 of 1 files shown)


Hamms:
I've discovered a potential problem; when doing final testing, I was occasionally running into a "too many connections" issue. There's probably a way I can keep the number of connections we're making under control, but it's going to be a pain to test because it seems to only happen when I'm testing with the full amount of data, which takes a _while_.

Hamms:
Okay, I _think_ I've confirmed that the too many connections error just comes from testing something over and over again in a short period of time, and won't happen on the actual sync. I will be keeping a close eye on the sync after this gets merged just to be sure, but that does mean that this PR is ready for review!

i18n/management/commands/publish_i18n.py
```diff
@@ -1,16 +1,45 @@
...
+    if not obj.should_be_translated:
+        return
+
+    pdf_generation_time = 0
+
+    for language_code in Command.language_codes:
+        translation.activate(language_code)
+        if hasattr(obj, 'publish'):
+            list(obj.publish(silent=True))
+        # Temporary hack to turn off PDF generation while still directing users to the
```
bethanyaconnor, line 28: nit: I think the comment is out of date and I missed removing it when I removed the "hack"

Hamms: ooooh, that makes sense. I was thinking that this code doesn't look like a hack

PR #303, opened 2020-07-29
add common sense education image to static assets

cforkish:
# Description

this simply adds the image we need to display for common sense education lessons to `static/img` so its url will be available for the subsequent work i'm doing hardcoding that image into those lessons.

common_sense_education

# Reviewer Checklist:

- [ ] Code is well-commented


cforkish:
good call @mvkski

PR #304, opened 2020-07-29
Hardcode Common Sense Logo

cforkish:
# Description

we need to hard-code the Common Sense Education logo above the overview for common sense lessons. we already have a sort of flag denoting those lessons, which is that they use a unique creative commons image. so rather than adding an additional flag to indicate the need for this hard-coded logo (which would need to stay in sync with the image choice), it was decided to use the image choice to determine the need for this logo.

Screen Shot 2020-07-28 at 5 04 13 PM

### Deployment strategy

as soon as this PR is deployed, i will manually remove the image from the markdown in all the common sense lessons.

## Links

- [jira](https://codedotorg.atlassian.net/browse/LP-1504)

## Testing story

# Reviewer Checklist:

- [ ] Code is well-commented

### DIFF
--- a/curricula/templates/curricula/partials/lesson_front.html
+++ b/curricula/templates/curricula/partials/lesson_front.html
@@ -26,6 +26,15 @@
 <div class="together overview-page row">
 
     <div class="col-sm-7 left-col">
+        {% comment %}
+        lessons using the CREATIVE_COMMONS_BY_NC_ND_IMAGE image also need
+        the Common Sense Education logo above the overview
+        {% endcomment %}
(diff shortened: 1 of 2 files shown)


curricula/templates/curricula/partials/lesson_front.html
```diff
@@ -26,6 +26,15 @@ <h4>
 <div class="together overview-page row">
 
     <div class="col-sm-7 left-col">
+        {% comment %}
+        lessons using the CREATIVE_COMMONS_BY_NC_ND_IMAGE image also need
```
clareconstantine, line 30: should the contents of the comment be indented?

cforkish: yes they should!

PR #307, opened 2020-08-27
more attempts to fix the S3 interaction for the i18n sync

Hamms:
The whole storage abstraction we have going on here is significantly more confusing than it needs to be.

So the way it works is the "absolute path" really only works for local filesystems; when we're using a virtual filesystem like we do in production, we need to use the relative path and also make sure to use the Storage-specific `open` operator rather than the OS-provided one.

We also need to be careful with how we handle creation of directories, since the whole concept of directories is itself filesystem specific and doesn't apply to S3.

# Reviewer Checklist:

- [ ] Code is well-commented


Hamms:
As an aside, it's _super_ annoying that heroku doesn't have a way of doing something like "spin up a one-off dyno based on a feature branch" so this can be tested without deploying

PR #309, opened 2020-09-03
Re-parallelize the publish operation

Hamms:
Also make it be a bit more explicit about exactly which languages it's publishing.

When we added the ability for the sync to skip publishing languages without translation changes in https://github.com/code-dot-org/curriculumbuilder/pull/305, we undid the parallelization stuff to get it to work. This PR reintroduces parallelization! We also update the time tracking and time reporting stuff to be a bit more explicit about exactly what's going on.

### DIFF
--- a/i18n/management/commands/publish_i18n.py
+++ b/i18n/management/commands/publish_i18n.py
@@ -73,10 +73,14 @@ class Command(BaseCommand):
         self.html_publishing_time += (end_time - start_time)
 
         if language_code in settings.LANGUAGE_GENERATE_PDF:
-            pdf_start_time = time.time()
-            self.pool.imap_unordered(publish_object_pdfs, objects.all())
-            pdf_end_time = time.time()
-            self.pdf_publishing_time += (pdf_end_time - pdf_start_time)
+            try:
+                pdf_start_time = time.time()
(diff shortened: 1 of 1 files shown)


i18n/management/commands/publish_i18n.py
```diff
@@ -23,46 +35,60 @@ class Command(BaseCommand):
...
+        # files. So, we just determine changes on a per-language (rather than per-language and also
+        # per-model) basis. This should be sufficient for our needs given the relatively low
+        # translation activity for this project, but we may need to make this more sophisticated in
+        # the future if that changes.
+        return len(self.changes.get(language_code, [])) > 0
+
+    def publish_objects_in_language(self, objects, language_code):
         translation.activate(language_code)
 
-        if hasattr(obj, 'publish'):
```
bethanyaconnor, line 46: Do we not need this check anymore?

bethanyaconnor (approved):

i18n/management/commands/publish_i18n.py
Hamms: we do, it just moved! It's up here now https://github.com/code-dot-org/curriculumbuilder/pull/309/files#diff-84e2554603b1b28645da9724b3100e34R20

PR #313, opened 2020-10-14
Use external source of truth for LANGUAGES setting

Hamms:
So that we can access this list of languages both from the CurriculumBuilder project (where we need it for the i18n sync) and from the Code.org project (where we need it for various display/linking operations)

## Testing story

Tested manually

# Reviewer Checklist:

- [ ] Code is well-commented

### DIFF
--- a/curriculumBuilder/settings.py
+++ b/curriculumBuilder/settings.py
@@ -1,6 +1,8 @@
 from __future__ import absolute_import, unicode_literals
 
+import boto3
 import socket
+import json
 
 import os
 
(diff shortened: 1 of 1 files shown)


Hamms:
Also see https://github.com/code-dot-org/code-dot-org/pull/37222 for where I added the necessary AWS permissions

davidsbailey (approved):
Nice solution @Hamms ! per offline conversation, my understanding is that this would only affect the i18n sync, and error checking will be added there to catch the case where the expected list of languages was not available here in CB.

PR #315, opened 2020-11-04
start dumping unit, chapter, lesson and activity data to json

davidsbailey:
This is the very start of exporting Curriculum Builder data as JSON so that we can import it into Code Studio. It sets up summarize methods on Unit, Chapter, Lesson and Activity so that we can dump some fields about a Unit and the objects in it.
Sample output (truncated):
```
(cb) Dave-MBP:~/src/cb (dump-unit-json)$ ./manage.py shell_plus
In [1]: u = Unit.objects.get(pk=5103)             
In [2]: import json                               
In [3]: print(json.dumps(u.summarize(), indent=4))
{
    "unit_name": "coursea-2020", 
    "chapters": [
        {
            "lessons": [
                {
                    "duration": "", 
                    "overview": "This lesson was originally created by Co...", 
                    "activities": [
                        {
                            "content": "### Key Vocabulary\r\n* **online**: using ...", 
                            "order": 0, 
                            "name": "Watch: My Online Neighborhood"
                        }, 
                        {
                            "content": "**Say**:\u00a0*Now we're going to learn a sho...", 
                            "order": 1, 
                            "name": "Read: Safety, It Rules!"
                        }, 
                        {
                            "content": "**Say**:\u00a0*The computer makes it possible...", 
                            "order": 2, 
                            "name": "Explore: Internet Field Trip"
                        }, 
                        {
                            "content": "**Say**:\u00a0*Today we talked about how the ...", 
                            "order": 3, 
                            "name": "Pause & Think Moment"
                        }, 
                        {
                            "content": "Here are additional resources you can pr...", 
                            "order": 4, 
                            "name": "Extended Learning"
                        }
                    ], 
                    "code_studio_url": ""
                }
            ], 
            "number": 1, 
            "questions": ""
        }, 
        {
            "lessons": [
                {
                    "duration": "", 
                    "overview": "This lesson will give students an idea o...", 
                    "activities": [
                        {
                            "content": "### Behaving in the Computer Lab\r\n\r\nGoal...", 
                            "order": 0, 
                            "name": "Warm Up"
                        }, 
                        {
                            "content": "Project a puzzle from lesson 2. Show the...", 
                            "order": 1, 
                            "name": "Preview Online Puzzles"
                        }, 
...
```

You could also use it to export an individual lesson as JSON:
```
In [4]: lesson = Lesson.objects.get(pk=5119)           
In [5]: print(json.dumps(lesson.summarize(), indent=4))
{
    "duration": "", 
    "overview": "In this online activity, students will h...", 
    "activities": [
        {
            "content": "### Introduction\r\n\r\nReview \"The Big Even...", 
            "order": 0, 
            "name": "Warm Up"
        }, 
        {
            "content": "This activity will help bring the unplug...", 
            "order": 1, 
            "name": "Bridging Activity - Events"
        }, 
        {
            "content": "### Online Puzzles and Free Play\r\n\r\nThis...", 
            "order": 2, 
            "name": "Main Activity"
        }, 
        {
            "content": "### Journaling\r\n\r\nHaving students write ...", 
            "order": 3, 
            "name": "Wrap Up"
        }, 
        {
            "content": "Use these activities to enhance student ...", 
            "order": 4, 
            "name": "Extended Learning"
        }
    ], 
    "code_studio_url": null
}
```
I got the pk numbers as follows:
1. sign in to codecurricula.com/admin
2. find a page like https://www.codecurricula.com/csf-20/coursea/1/
3. press `-->` at top left, and then click `lesson admin`
4. grab the number at the end of the url that opens: https://www.codecurricula.com/admin/lessons/lesson/5105/

a similar process works for units.

### DIFF
--- a/curricula/models.py
+++ b/curricula/models.py
@@ -689,11 +689,12 @@ class Unit(InternationalizablePage, RichText, CloneableMixin, Ownable):
 
         return duplicate
 
-    def to_json(self):
+    def summarize(self):
         summary = {
-            'unit_name': self.unit_name
+            'unit_name': self.unit_name,
+            'chapters': map(lambda ch: ch.summarize(), self.chapters)
         }
-        return json.dumps(summary)
+        return summary
 
 
 """
@@ -820,6 +821,13 @@ class Chapter(InternationalizablePage, RichText, CloneableMixin, Ownable):
         duplicate.save()
 
         return duplicate
+
+    def summarize(self):
+        summary = {
+            'number': self.number,
+            'questions': self.questions
+        }
+        return summary
 
 
 """
--- a/curricula/models.py
+++ b/curricula/models.py
@@ -825,7 +825,8 @@ class Chapter(InternationalizablePage, RichText, CloneableMixin, Ownable):
     def summarize(self):
         summary = {
             'number': self.number,
-            'questions': self.questions
+            'questions': self.questions,
+            'lessons': map(lambda l: l.summarize(), self.lessons)
         }
         return summary
 
--- a/curricula/models.py
+++ b/curricula/models.py
@@ -689,6 +689,12 @@ class Unit(InternationalizablePage, RichText, CloneableMixin, Ownable):
 
         return duplicate
 
+    def to_json(self):
+        summary = {
+            'unit_name': self.unit_name
+        }
+        return json.dumps(summary)
+
 
 """
 Unit Chapter
--- a/lessons/models.py
+++ b/lessons/models.py
@@ -732,6 +732,14 @@ class Lesson(InternationalizablePage, RichText, CloneableMixin, Filterable):
         for standard in self.standards.all():
             results[standard.framework.slug][standard.top_category.shortcode].append(standard)
 
+    def summarize(self):
+        summary = {
+            'overview': self.overview,
+            'duration': self.duration,
+            'code_studio_url': self.code_studio_url,
+        }
+        return summary
+
 
 """
 Activities that compose a lesson
--- a/lessons/models.py
+++ b/lessons/models.py
@@ -733,10 +733,12 @@ class Lesson(InternationalizablePage, RichText, CloneableMixin, Filterable):
             results[standard.framework.slug][standard.top_category.shortcode].append(standard)
 
     def summarize(self):
+        activities = self.activity_set.iterator()
         summary = {
             'overview': self.overview,
             'duration': self.duration,
             'code_studio_url': self.code_studio_url,
+            'activities': map(lambda a: a.summarize(), activities)
         }
         return summary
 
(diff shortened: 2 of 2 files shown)


bethanyaconnor (approved):
Looks great! Thanks for doing this!

davidsbailey:
I was starting to work on the next PR to add a URL / route for this, when I discovered that this already exists: https://www.codecurricula.com/metadata/coursea-2020.json
Screen Shot 2020-11-04 at 10 01 42 AM

Some code snippets:
https://github.com/code-dot-org/curriculumbuilder/blob/50b907c354312aab8deaddc06e3bd537ea2a7b3b/curricula/urls.py#L24
https://github.com/code-dot-org/curriculumbuilder/blob/50b907c354312aab8deaddc06e3bd537ea2a7b3b/curricula/views.py#L1002-L1002
https://github.com/code-dot-org/curriculumbuilder/blob/50b907c354312aab8deaddc06e3bd537ea2a7b3b/curricula/views.py#L1002-L1009
https://github.com/code-dot-org/curriculumbuilder/blob/50b907c354312aab8deaddc06e3bd537ea2a7b3b/curricula/serializers.py#L86-L93

@dmcavoy (for tomorrow) do you have any knowledge of how this route is/was used? If unused, it could be a bit better to add to it rather than creating a separate parallel codepath.

davidsbailey:
OK, this turned out to be easy to find. This appears to be used by the part of the levelbuilder script edit page that is already set up to pull over stage/lesson descriptions from CB: 
https://github.com/code-dot-org/code-dot-org/blob/4db34f7d409273b0c8d36f35c20a5e5f2c2ee046/apps/src/lib/levelbuilder/script-editor/LessonDescriptions.jsx#L103-L111

This seems like it is doing something really similar to what we want to do, and there is already a bit of logic to match CB lessons to LB lessons, so I think we do want to _consider_ building on top of this. we would have to be extra careful not to change any existing behavior, to make sure that clicking these buttons on existing scripts does not mess anything up:
Screen Shot 2020-11-04 at 10 16 51 AM
Screen Shot 2020-11-04 at 10 16 59 AM

Going this route would kind of wipe out a lot of what we drew on the whiteboard... though it does share the general idea of getting CB data into json format to be consumed by code studio. Anyway it seems like it could be a big shortcut so I am going to look into it a bit more and consider abandoning this PR.

davidsbailey:
Sorry for the unfiltered thinking here, but now I am thinking that the script edit page JS is not a great place to do the merge, because it doesn't know to save activities, etc. So it might make sense to share the CB route which produces the JSON, but probably not the mechanism for merging it into Code Studio.

bethanyaconnor:
The json that's currently being dumped doesn't seem to have any activity data, so we'd still need to build _something_ for activities.

davidsbailey:
Agreed. Serializers look better than what I have here, so I think it's down to adding a new set of serializers, or adding activity info to existing serializers.

PR #318, opened 2020-11-05
serialize Unit, Chapter, Lesson and Activity using serializers

davidsbailey:
Reimplements #315 using serializers. This seems a little less error-prone, for example switching over to serializers gave me a friendly error when a field I asked for did not exist. 

I decided to make a separate set of serializers for now, because (1) I want to be able to keep junk we don't need out of this json, and (2) I don't want to clutter up the json we already use for pulling stage details at https://www.codecurricula.com/metadata/coursea-2020.json , and (3) this gives extra insurance against somehow breaking the process of pulling stage details.

Current usage:
```
(cb) Dave-MBP:~/src/cb (serialize-activities)$ ./manage.py shell_plus
In [1]: from curricula.serializers import *
In [2]: import json
In [3]: unit = Unit.objects.get(unit_name='coursea-2020')
In [4]: data = UnitExportSerializer(unit).data
In [5]: print(json.dumps(data, indent=4))     
```
Sample output: [coursea-2020.json.zip](https://github.com/code-dot-org/curriculumbuilder/files/5491110/coursea-2020.json.zip)

A similar approach could be used to dump an individual lesson.

The next PR will add a route to retrieve the result of this operation as json, similar to https://www.codecurricula.com/metadata/coursea-2020.json .

### DIFF
--- a/curricula/serializers.py
+++ b/curricula/serializers.py
@@ -5,6 +5,9 @@ from curricula.models import Unit, Curriculum
 from lessons.models import Lesson, Resource, Vocab, Annotation, Standard
 from documentation.models import Block
 
+"""
+Curriculum serializers
+"""
 
 class ResourceSerializer(serializers.ModelSerializer):
     html = serializers.SerializerMethodField()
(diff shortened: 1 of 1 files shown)


bethanyaconnor (approved):
Played with this locally and looks good!

PR #319, opened 2020-11-05
add url to export unit for import into code studio

davidsbailey:
#318 serialized unit / chapter / lesson / activity. This PR adds a url to download the result as json.

Screen Shot 2020-11-04 at 7 25 26 PM

### DIFF
--- a/curricula/views.py
+++ b/curricula/views.py
@@ -1009,6 +1009,19 @@ def unit_element(request, unit_name, format=None):
     serializer = UnitSerializer(unit)
     return Response(serializer.data)
 
+
+@api_view(['GET', ])
+def unit_export(request, unit_name, format=None):
+    try:
+        unit = Unit.objects.get(login_required=False, status=2, unit_name=unit_name)
+    except MultipleObjectsReturned:
+        logger.exception("Warning - found multiple units referencing the unit %s" % unit_name)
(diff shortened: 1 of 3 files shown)


curricula/views.py
```diff
@@ -1009,6 +1009,19 @@ def unit_element(request, unit_name, format=None):
     serializer = UnitSerializer(unit)
     return Response(serializer.data)
 
+
+@api_view(['GET', ])
+def unit_export(request, unit_name, format=None):
+    try:
+        unit = Unit.objects.get(login_required=False, status=2, unit_name=unit_name)
+    except MultipleObjectsReturned:
```
bethanyaconnor, line 1017: Should we do anything for no objects returned? Or are we okay with that just failing?

bethanyaconnor (approved):

curricula/views.py
davidsbailey: good idea! it looks like there is a helper for that. before:

Screen Shot 2020-11-05 at 2 55 10 PM

after:

Screen Shot 2020-11-05 at 2 54 24 PM

docs: https://docs.djangoproject.com/en/1.8/topics/http/shortcuts/#get-object-or-404

davidsbailey: I will fix up the code studio script to give some idea that this 404 has happened in this case.

PR #322, opened 2020-11-10
include chapter title and description when exporting cb units

davidsbailey:


bethanyaconnor (approved):

davidsbailey:
added one more field needed for import into code studio.

PR #324, opened 2020-11-13
export show_calendar field on Unit object

davidsbailey:
needed for https://github.com/code-dot-org/code-dot-org/pull/37845

### DIFF
--- a/curricula/serializers.py
+++ b/curricula/serializers.py
@@ -132,7 +132,7 @@ class UnitExportSerializer(serializers.ModelSerializer):
 
     class Meta:
         model = Unit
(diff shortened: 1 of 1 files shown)


bethanyaconnor (approved):

dmcavoy:
We don't currently have a place to save this information yet. Should I work to set that up?

davidsbailey:
> We don't currently have a place to save this information yet. Should I work to set that up?

https://github.com/code-dot-org/code-dot-org/pull/37845 does the work to save this information in code studio. we do not need to save this information into .script files, because only non-legacy scripts will use this field. There shouldn't be any explicit work needed for this in the new seeding system. We would need future work to actually show this calendar, which appears to be tracked by https://codedotorg.atlassian.net/browse/PLAT-243 .  I think that covers it, but please let me know if there is any other work I am missing!

PR #329, opened 2020-12-04
Adding the 'slug' attribute for Lessons in the Admin panel

daynew:
# Description
Adding the `slug` attribute to the Admin panel for the Lesson model. This is helpful because currently the `slug` is used by the i18n system as the "key" for looking up the Lesson's translations. Having this field available for viewing and editing will enable us to quickly debug issues where different lessons accidentally have the same/wrong `slug`.

## Testing story
* Manually tested on localhost
  * http://localhost-studio.code.org:8000/admin/lessons/lesson/2640/

## Screenshots
image

# Reviewer Checklist:

- [ ] Code is well-commented

### DIFF
--- a/lessons/admin.py
+++ b/lessons/admin.py
@@ -328,7 +328,7 @@ class LessonAdmin(PageAdmin, AjaxSelectAdmin, CompareVersionAdmin, FilterableAdm
         fieldsets = (
             (None, {
                 'fields': ['title', 'short_title', status_fields, 'image',
-                           'overview', 'keywords', ('description', 'gen_description'), 'code_studio_url'],
+                           'overview', 'slug', 'keywords', ('description', 'gen_description'), 'code_studio_url'],
             }),
             ('Assessment', {
                 'fields': ['assessment'],


dmcavoy:
In the example given the 'plugged' part of the url comes from the unit and is set on the unit page. So I think the slug should just be 'dance-party'.

I do wonder if having this field will confuse editors who usually rely on lesson numbering as the slug

daynew:
> In the example given the 'plugged' part of the url comes from the unit and is set on the unit page. So I think the slug should just be 'dance-party'.
> 
> I do wonder if having this field will confuse editors who usually rely on lesson numbering as the slug

I don't know what the value of a slug "should" be. This PR is meant to simply expose what the value actually is and also modify it if there is something wrong with it.

A translation issue I have been debugging seems related to how Lessons have been copied. The `slug` attribute has been blindly copied so now we have an "AI For Oceans" Lesson which has the slug `plugged/intro-to-app-lab`. This PR will enable us to quickly go in and change it to `plugged/ai-for-oceans` or whatever it "should" be.

dmcavoy:
> > In the example given the 'plugged' part of the url comes from the unit and is set on the unit page. So I think the slug should just be 'dance-party'.
> > I do wonder if having this field will confuse editors who usually rely on lesson numbering as the slug
> 
> I don't know what the value of a slug "should" be. This PR is meant to simply expose what the value actually is and also modify it if there is something wrong with it.
> 
> A translation issue I have been debugging seems related to how Lessons have been copied. The `slug` attribute has been blindly copied so now we have an "AI For Oceans" Lesson which has the slug `plugged/intro-to-app-lab`. This PR will enable us to quickly go in and change it to `plugged/ai-for-oceans` or whatever it "should" be.

Ya I think my concern is the field will also show to the curriculum and PL team and possibly be confusing for them.

PR #333, opened 2020-12-18
Fixing the parallel processing in the publish_i18n step

daynew:
*Bug* - The `publish_i18n` step of the i18n sync in CurriculumBuilder is not producing and translated files.
*Cause* - `pool.imap_unordered` is not a synchronous call, so work was getting queued up but the code never waited for the work to finish before ending the script.
*Fix* - Use `pool.map` which is a synchronous call. The translated content for a language will be completed before moving onto the next language.

## Testing story
* Configured `publish_i18n` to write the HTML files to my local disk and confirmed that there are now translations in all the languages.

# Reviewer Checklist:

- [ ] Code is well-commented

### DIFF
--- a/i18n/management/commands/publish_i18n.py
+++ b/i18n/management/commands/publish_i18n.py
@@ -37,7 +37,6 @@ class Command(BaseCommand):
         self.total_elapsed_time = 0
         self.pdf_publishing_time = 0
         self.html_publishing_time = 0
-        self.pool = multiprocessing.Pool(multiprocessing.cpu_count())
 
         with open(CHANGES_JSON) as changes_json:
             self.changes = json.load(changes_json)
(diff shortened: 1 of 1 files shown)


i18n/management/commands/publish_i18n.py
```diff
@@ -59,29 +58,35 @@ def has_changes(self, language_code):
...
-
         # By default, the spawned processes will attempt to reuse the existing database connections,
         # which will cause conflicts. To prevent this, we explicitly close all database connections
         # immediately before spawning the processes, which will force them to each open their own
         # connection.
         db.connections.close_all()
 
+        translation.activate(language_code)
+
+        pool = multiprocessing.Pool(multiprocessing.cpu_count())
```
annaxuphoto, line 69: what's the advantage of creating a separate `pool` variable vs `self.pool` ?

daynew: I made this change to guarantee the child process inherited the parent process state after `translation.activate(language_code)` is called. Its probably true that I could revert `pool` back to a instance variable, but being an instance variable seems unnecessary anyways because no other class methods are using `pool`.

annaxuphoto (approved):

PR #337, opened 2021-01-22
Remove static i18n directory

annaxuphoto:
# Description

Problem summary: the CB i18n-sync reads from the` i18n/static/translations` directory, but it writes to the `i18n/translations` directory. Thus, CB translations have not been updated since August 2020.

This change removes the `static/` directory from the sync source directory and locale paths so that CB reads from the correct location

## Links

- [spec]()
- [jira](https://codedotorg.atlassian.net/secure/RapidBoard.jspa?rapidView=28&modal=detail&selectedIssue=FND-1328)

## Testing story

Tested the sync-down script both locally and to the `cdo-curriculum-devel` S3 bucket to ensure correct upload. Verified locally that new translations show up (UI element translations are the most obvious giveaway)

Before:
Screen Shot 2021-01-22 at 1 18 46 PM

After:
Screen Shot 2021-01-22 at 1 19 03 PM

# Reviewer Checklist:

- [ ] Code is well-commented


i18n/utils.py
```diff
@@ -25,10 +25,6 @@ class I18nFileWrapper:
     def i18n_dir(cls):
         return os.path.dirname(__file__)
```
Hamms, line 26: I believe this should be something like

```suggestion
        return settings.I18N_DIR
```

Hamms (commented):
This looks like a great start! It's not clear to me that we want to standardize around the `i18n` directory rather than the `i18n/static` directory, though; in particular, this means that we'll be interacting with these files directly in https://github.com/code-dot-org/curriculumbuilder/tree/master/i18n rather than in a subdirectory. At the very least, it means that https://github.com/code-dot-org/curriculumbuilder/blob/master/i18n/.gitignore will need to be updated, but more generally it represents a level of blending of code and data that I don't love

annaxuphoto:
My understanding is that the translations will still be in `i18n/translations`, is that a sufficient differentiation between code/data?

Hamms:
the translations will be in `i18n/translations`, but the source strings will be gathered into `i18n/source` and then redacted into `i18n/upload`, which is more noise than I'd prefer

PR #338, opened 2021-01-22
setting storage location to i18n/static by default

annaxuphoto:
# Description

Problem summary: the CB i18n-sync reads from the` i18n/static/translations` directory, but it writes to the `i18n/translations` directory. Thus, CB translations have not been updated since August 2020.

This change adjusts the default storage location to `i18n/static` instead of `i18n` so that the sync-down uploads files to the same directory that CB reads translations from

## Links

- [spec]()
- [jira](https://codedotorg.atlassian.net/secure/RapidBoard.jspa?rapidView=28&modal=detail&selectedIssue=FND-1328)

## Testing story

Tested the sync-down script both locally and to the `cdo-curriculum-devel` S3 bucket to ensure correct upload. Verified locally that new translations show up (UI element translations are the most obvious giveaway)

Before:
Screen Shot 2021-01-22 at 1 18 46 PM

After:
Screen Shot 2021-01-22 at 1 19 03 PM

# Reviewer Checklist:

- [ ] Code is well-commented

### DIFF
--- a/curriculumBuilder/settings.py
+++ b/curriculumBuilder/settings.py
@@ -14,8 +14,7 @@ from django.utils.translation import ugettext_lazy as _
 from urlparse import urlparse
 
 BASE_DIR = os.path.dirname(os.path.dirname(__file__))
-I18N_DIR = os.path.join(BASE_DIR, "i18n")
-I18N_STATIC_DIR = os.path.join(I18N_DIR, "static")
+I18N_STORAGE_LOCATION = os.path.join("i18n", "static")
 
 ######################
 # MEZZANINE SETTINGS #
@@ -164,7 +163,7 @@ LANGUAGE_GENERATE_PDF = (
 )
 
 LOCALE_PATHS = (
-    os.path.join(I18N_STATIC_DIR, 'translations/'),
+    os.path.join(BASE_DIR, I18N_STORAGE_LOCATION, 'translations/'),
 )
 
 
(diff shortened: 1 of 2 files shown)


daynew (approved):

curriculumBuilder/settings.py
```diff
@@ -14,7 +14,8 @@
 from urlparse import urlparse
 
 BASE_DIR = os.path.dirname(os.path.dirname(__file__))
-I18N_STATIC_DIR = os.path.join(BASE_DIR, "i18n", "static")
+I18N_DIR = os.path.join(BASE_DIR, "i18n")
+I18N_STATIC_DIR = os.path.join(I18N_DIR, "static")
```
Hamms, line 18: I don't think this is what we want; I'd expect this to end up being a value like `/path/to/the/project/curriculumBuilder/i18n/static`

i18n/utils.py
```diff
@@ -94,5 +94,5 @@ def storage(cls):
         if cls._storage is None:
             storage = getattr(settings, 'I18N_STORAGE', 'django.core.files.storage.FileSystemStorage')
             storage_cls = import_string(storage)
-            cls._storage = storage_cls(location=getattr(settings, 'I18N_STORAGE_LOCATION', 'i18n'))
+            cls._storage = storage_cls(location=getattr(settings, 'I18N_STORAGE_LOCATION', 'i18n/static'))
```
Hamms, line 97: I'd suggest we have a goal of entirely getting rid of either `I18N_STATIC_DIR` or of `I18N_STORAGE_LOCATION`, rather than just updating them both to incidentally have the same values

Hamms (changes requested):

curriculumBuilder/settings.py
annaxuphoto: Do we want this to be a relative path?

Hamms: I believe that the `i18n_dir` and `static_dir` helper methods are used by the I18nFileWrapper; since they're now referencing these settings, the settings need to be compatible with non-local-filesystem storage solutions

i18n/utils.py
```diff
@@ -23,7 +23,10 @@ class I18nFileWrapper:
 
     @classmethod
     def i18n_dir(cls):
-        return os.path.dirname(__file__)
+        static = "/static"
+        if static in cls.storage().location:
+            return cls.storage().location.replace(static, "")
+        return cls.storage().location
```
Hamms, line 29: I believe this represents a bit of a misunderstanding as to what the return value of this method represents. I'd recommend you perhaps start by adding a docstring for this method and also for `static_dir` below; take some time to look around and see how they're used and then explicitly document the intended purpose of this method based on those observations. That should make it much more clear how to think about this kind of implementation tweak

PR #341, opened 2021-02-06
adding python script to locally download django.po and .mo files in CB

annaxuphoto:
# Description
As discussed offline, we're adding a script that downloads `django.po`  files from Crowdin and compiles them into their respective `django.mo` files. The script is very similar to the [`download_translations`](https://github.com/code-dot-org/curriculumbuilder/blob/master/i18n/management/crowdin.py#L136) function in crowdin.py.

Functionality was tested by: 
1. removing all translations from my local file system, running the `i18n_message_file_sync` command, 
2. verifying that the `django.po` and `.mo` files have been successfully downloaded to the `i18n/static/translations` directory, and
3. verifying that the translated UI element strings show correctly on the local server
Before:
Screen Shot 2021-02-05 at 4 52 25 PM
After:
Screen Shot 2021-02-05 at 4 52 31 PM

## Links

- [spec]()
- [jira]()

## Testing story

# Reviewer Checklist:

- [ ] Code is well-commented

### DIFF
--- a/i18n/management/commands/i18n_message_file_sync.py
+++ b/i18n/management/commands/i18n_message_file_sync.py
@@ -1,0 +1,86 @@
+# pylint: disable=missing-docstring
+import glob
+import os
+import json
+
+from django.core import management
+from django.core.management.base import BaseCommand
+from django.utils.translation import to_locale
+from django.core.files.storage import FileSystemStorage
+
+from i18n.management.crowdin import Crowdin
+from i18n.management.utils import log, get_non_english_language_codes
+from i18n.utils import I18nFileWrapper
+
+
+class Command(BaseCommand):
+
+    # Retrieves all django.po files from Crowdin and
+    # saves them locally
+    def download_message_file(self):
+        # Download changes!
+        for language_code in get_non_english_language_codes():
+
+            locale = to_locale(language_code)
+            language_dir = I18nFileWrapper.locale_dir_absolute(locale)
+            if not os.path.exists(language_dir):
+                os.makedirs(language_dir)
+
+            # Load existing etags from previous sync, if it exists.
+            # Note that here we're using locale_dir and not locale_dir_absolute; we want to
+            # specifically retrieve the etags from wherever they are persisted, even if that's not
+            # the local filesystem.
+            etags = {}
+            etags_path = os.path.join(I18nFileWrapper.locale_dir(locale), "crowdin_etags.json")
+            if I18nFileWrapper.storage().exists(etags_path):
+                log("loading existing etags from %s" % etags_path)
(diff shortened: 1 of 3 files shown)


Hamms:
We did discuss this offline, so I am recording my thoughts from that discussion here. I think this is a very bad idea.

Hamms (commented):
Is this not already handled by the `download_translations` method you mention? If not, we should simply update that method to include these. We don't want to treat the `.po` files any differently from the `.json` files in the sync process.

i18n/management/commands/i18n_message_file_sync.py
```diff
@@ -0,0 +1,86 @@
...
+from i18n.management.utils import log, get_non_english_language_codes
+from i18n.utils import I18nFileWrapper
+
+
+class Command(BaseCommand):
+
+    # Retrieves all django.po files from Crowdin and
+    # saves them locally
+    def download_message_file(self):
+        # Download changes!
```
bethanyaconnor, line 21: A lot of this code is duplicated from download_translations. Is there a way to either parameterize that function or extract some of the duplicate logic?

annaxuphoto: For sure, I'll update the function with an optional argument

i18n/management/commands/i18n_message_file_sync.py
```diff
@@ -5,81 +5,16 @@
...
-            if isinstance(I18nFileWrapper.storage(), FileSystemStorage) and not os.path.exists(storage_etags_path):
-                os.makedirs(storage_etags_path)
-            with I18nFileWrapper.storage().open(etags_path, 'w') as etags_file:
-                json.dump(etags, etags_file, sort_keys=True, indent=4)
-
     def handle(self, *args, **options):
         log("Checking for updates in django.po")
         try:
-            self.download_message_file()
+            Crowdin().download_translations(["django.po"])
```
Hamms, line 17: I'm completely confused by this. The download translations method already downloads this file, why are we doing it again?

annaxuphoto: I meant this script to be run locally as an easy way to gather any changes in `django.po` that need to be checked in - it's not a part of the i18n sync

Hamms: ahhh, so this _is_ in fact part of the work that I think we should not be doing. That had been my initial impression, but I second-guessed that

annaxuphoto: haha, yep

bethanyaconnor (approved):
Thanks for adding that parameter -- the code looks a lot simpler now! This looks good to me

PR #342, opened 2021-02-06
adding message files for CB

annaxuphoto:
# Description
Checking in `django.po` and `django.mo` message files for Curriculum Builder UI translated elements

## Links

- [spec]()
- [jira]()

## Testing story

# Reviewer Checklist:

- [ ] Code is well-commented


bethanyaconnor (approved):
Do you actually need to check in the .po files? I could see reasons to still check them in, just curious

annaxuphoto:
We don't need it, but I'd like to have them for human-readability

bethanyaconnor:
Makes sense!

Hamms:
I will reiterate that I do not think we should be storing serialized data files in git

PR #343, opened 2021-02-10
upgrade to psycopg2 version 2.8.6

davidsbailey:
This was the first step for me in fixing a long list of build problems on OS X. Here is everything that i did to get my system working:

1. at some point, brew made me upgrade to openssl@1.1.

2. `pip install -r requirements.txt` broke with:
```
django.core.exceptions.ImproperlyConfigured: Error loading psycopg2 module: dlopen(/Users/dsb/.virtualenvs/cb/lib/python2.7/site-packages/psycopg2/_psycopg.so, 2): Library not loaded: /usr/local/opt/openssl/lib/libssl.1.0.0.dylib
```
I tried installing an older version of openssl, but that didn't work. 

3. I upgraded psycopg2 to the latest version, as shown in this PR. 

4. `pip install -r requirements.txt` now succeeds

5. `debug=true python manage.py runserver_plus` fails with:
```
OperationalError: could not connect to server: No such file or directory
	Is the server running locally and accepting
	connections on Unix domain socket "/tmp/.s.PGSQL.5432"?
```

6. postgres appeared to not be running:
```
(cb) Dave-MBP:~/src/cb (master *)$ ps -ef | grep postgres
  501 20433 12580   0  6:53AM ttys004    0:00.00 grep postgres
```

7. from https://stackoverflow.com/questions/13410686/postgres-could-not-connect-to-server, I ran:
```
(cb) Dave-MBP:~/src/cb (master *)$ postgres -D /usr/local/var/postgres
2021-02-10 07:02:13.089 PST [21823] FATAL:  database files are incompatible with server
2021-02-10 07:02:13.089 PST [21823] DETAIL:  The data directory was initialized by PostgreSQL version 11, which is not compatible with this version 13.1.
```

8. from https://stackoverflow.com/questions/17822974/postgres-fatal-database-files-are-incompatible-with-server, I ran:
```
brew postgresql-upgrade-database
```

9. `brew services restart postgresql`

10. success!
```
(cb) Dave-MBP:~/src/cb (master *)$ debug=true python manage.py runserver_plus
 * Running on http://127.0.0.1:8000/ (Press CTRL+C to quit)
```

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -71,7 +71,7 @@ newrelic==2.96.0.80
 oauthlib==1.0.3
 pdfkit==0.5.0
 plotly==1.9.3
-psycopg2==2.8.4
+psycopg2==2.8.6
 pycurl==7.43.0.3
 pyPdf==1.13
 pytz==2015.7


davidsbailey:
@dmcavoy @bethanyaconnor @Hamms no rush, but it would be great if you could check that this change will not break you, and/or if the steps might fix any CB build problems.

bethanyaconnor:
Hmm I used curriculum builder locally yesterday and did have a similar problem to your step 5, though I resolved it a different way. My openssl version is `openssl@1.1 1.1.1i`.

I fetched this branch and tried to run `pip install -r requirements.txt` but it crashed when trying to upgrade psycopg2. I'll dig a bit more to see if it's just my environment or if there's an easy fix

bethanyaconnor:
Okay, following instructions from [stack overflow](https://stackoverflow.com/questions/26288042/error-installing-psycopg2-library-not-found-for-lssl) I ran `env LDFLAGS="-I/usr/local/opt/openssl/include -L/usr/local/opt/openssl/lib" pip install -r requirements.txt` instead and step 2 finished successfully. But after that my server ran successfully...

I resolved the error message from step 5 yesterday by loosely following the instructions at https://stackoverflow.com/questions/42653690/psql-could-not-connect-to-server-no-such-file-or-directory-5432-error. Really, I just ran `rm /usr/local/var/postgres/postmaster.pid`

davidsbailey:
Thank you, @bethanyaconnor ! I noticed that I have `LDFLAGS` set to something similar in my .profile, but I don't see any mention of this in README.md or in our main repo's SETUP.md, so I imagine other folks will be likely to run into this when they upgrade psycopg2.

I'm not quite sure how best to address this, because my LDFLAGS are slightly different. Perhaps if we can determine a consistent value for each platform, then we can add them to the OS-specific sections of the readme.

davidsbailey:
```
(cb) Dave-MBP:~/src/cb (upgrade-psycopg2)$ export | grep FLAGS
CPPFLAGS=-I/usr/local/opt/openssl@1.1/include
LDFLAGS=-L/usr/local/opt/openssl@1.1/lib
(cb) Dave-MBP:~/src/cb (upgrade-psycopg2)$ ls -l /usr/local/opt/openssl*
lrwxr-xr-x  1 dsb  admin  28 Feb 10 06:31 /usr/local/opt/openssl@1.0 -> ../Cellar/openssl@1.0/1.0.2t
lrwxr-xr-x  1 dsb  admin  28 Jan  5 20:50 /usr/local/opt/openssl@1.1 -> ../Cellar/openssl@1.1/1.1.1i
```

bethanyaconnor:
I don't have those flags set but here is my ssl locations:
```(cb) bethanys-mbp:curriculumbuilder bethanyconnor$ ls -l /usr/local/opt/openssl*
lrwxr-xr-x  1 bethanyconnor  admin  28 Dec 10 09:25 /usr/local/opt/openssl -> ../Cellar/openssl@1.1/1.1.1i
lrwxr-xr-x  1 bethanyconnor  admin  28 Dec 10 09:25 /usr/local/opt/openssl@1.1 -> ../Cellar/openssl@1.1/1.1.1i```

It seems like I could use the same flags as you and I believe it would still work

Hamms:
Interesting. I've got openssl `1.1.1` installed, and psycopg2 works fine for me both with the old version and with this new one.

Hamms (approved):

bethanyaconnor (approved):
oops sorry -- didn't realize I hadn't approved!

davidsbailey:
Thanks all!

PR #349, opened 2021-03-15
include standards in lesson export

davidsbailey:
Continues on [PLAT-751](https://codedotorg.atlassian.net/browse/PLAT-751). For more background, see [Lesson Standards design doc](https://docs.google.com/document/d/1zvdv7y8VuRz8KqT1rzqeJAmvirCvUr1jS8OWy0p9FHk/edit)

There is an existing way to export lesson-standards relationships, but it has to be run manually for each script via web UI. This PR builds on the existing lesson/unit export functionality instead, which seems better because the existing import_unit_details.rb script is already set up to import this format and also serialize it to script json.


bethanyaconnor (approved):

davidsbailey:
Screen Shot 2021-03-15 at 10 27 27 AM

PR #359, opened 2021-04-27
Java updates for Heroku 18

Hamms:
The upgrade to Heroku 18 broke the i18n sync, because our hack to get the crowdin CLI to work in the Heroku 16 environment was, well, a hack, so it broke. Fortunately, Heroku 18 also fixed the weird pathing and dependency issues that forced us to go with the hack in the first place, so we should be able to replace it with something more standard!

Unfortunately, because heroku is heroku, I haven't been able to actually test this change out. It represents my best understanding of what's going on, but I may need to deploy a follow-up change if there are any unexpected problems.

# Reviewer Checklist:

- [ ] Code is well-commented


annaxuphoto:
Yay for getting around this hack!

annaxuphoto (approved):

PR #375, opened 2021-11-22
Update README.md

tim-dot-org:
Some issues i ran into setting the site up on my Windows WSL Ubuntu 20.04 laptop

I had some issues around the postgresql version, but I forget what i did to resolve that... :/


davidsbailey (approved):
perfect, thank you!

